PR #69, opened 2020-08-26
Add MailHog Service

r2luna:
It installs [MailHog](https://github.com/mailhog/MailHog)

Using [Valet Proxy](https://laravel.com/docs/master/valet#proxying-services) you can add a custom domain to MailHog.

```
valet proxy mailhog http://127.0.0.1:8025
```

Now you can use http://mailhog.test

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ But you can also easily enable ElasticSearch, Postgres, MSSQL, Redis, and more, 
 - MeiliSearch
 - Redis
 - Memcached
+- MailHog
 
 ## Requirements
 
--- a/app/Services/MailHog.php
+++ b/app/Services/MailHog.php
@@ -2,7 +2,7 @@
 
 namespace App\Services;
 
-class Mailhog extends BaseService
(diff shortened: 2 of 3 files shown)


mattstauffer:
Thanks @r2luna!

mattstauffer:
@r2luna I see that you didn't add a volume; does that mean Mailhog's Docker doesn't need a volume for any reason?

mattstauffer:
@r2luna Also--using 1025 as the default port.. that port is occupied on my Mac and I've never set this up with anything in particular. Is 1025 actually going to work for most people?

mattstauffer:
@josecanhelp Showed me 1025 is the right default:

image

So we'll just allow it to stay that way.

Looks like this is what's occupying mine, in case anyone else has this issue:

https://apple.stackexchange.com/questions/356622/what-is-identityservicesd

r2luna:
Great. I was about to send you that. About the volume, I'll send you another PR with the configuration for that.

Issue #68, opened 2020-08-26
Restart docker images after computer/docker restart


likeadeckofcards:
What is the best way to restart docker images after a computer/docker restart?

tcampbPPU:
For traditional docker images spun up should be able to add the additional policy via `$ docker update`
So, for example, list your docker images
```
$ docker ps -a
```

Then obtain the container ID for the image you want to update  and update its policy
```
$ docker update --restart unless-stopped contianer_id
```
Also, make sure you have the setting "Start Docker Desktop when you log in" enabled

bandgeekndb:
Would this make sense to add as a prompt when enabling a service? Ask if the user wants the service restarted with docker/computer and then run this command behind the scenes if it's desired?

mattstauffer:
Hm, I'm also wondering if it's possible to do that without the prompt. Problem is, if we auto restarted *every* container every time, some folks might not like that since they don't want all their containers always running, so we would have to keep track at any given moment of which are started and stopped, and that *also* seems like a pain...

josecanhelp:
This feels like a configuration that Takeout should not be managing. If a container needs to be started, just list the containers and select the ones you need to start. This will be a much smoother process in the node port, btw.

PR #71, opened 2020-08-27
Add command to restart stopped containers and stop running containers

infernobass7:
Add commands for users that are not familiar with docker to start a stopped container as well as stop a running container.

issue #68

### DIFF
--- a/app/Commands/HelpCommand.php
+++ b/app/Commands/HelpCommand.php
@@ -14,8 +14,10 @@ class HelpCommand extends Command
         'enable {service}' => 'Enable the provided service',
         'disable' => 'Disable a service from a list of options',
         'disable {service}' => 'Disable the provided service',
-        'restart' => 'Restart a stopped container',
-        'restart {container}' => 'Restart a stopped container',
+        'start' => 'Start a stopped container',
+        'start {container}' => 'Start a stopped container',
+        'stop' => 'Stop a running container',
+        'stop {container}' => 'Stop a running container',
         'list' => 'List all enabled services',
     ];
 
--- a/app/Commands/HelpCommand.php
+++ b/app/Commands/HelpCommand.php
@@ -14,10 +14,10 @@ class HelpCommand extends Command
         'enable {service}' => 'Enable the provided service',
         'disable' => 'Disable a service from a list of options',
         'disable {service}' => 'Disable the provided service',
-        'start' => 'Start a stopped container',
-        'start {container}' => 'Start a stopped container',
-        'stop' => 'Stop a running container',
-        'stop {container}' => 'Stop a running container',
+        'start' => 'Start a stopped container from a list of options',
+        'start {container}' => 'Start the provided stopped container (by ID)',
+        'stop' => 'Stop a running container from a list of options',
+        'stop {container}' => 'Stop the provided running container (by ID)',
         'list' => 'List all enabled services',
     ];
 
--- a/app/Commands/HelpCommand.php
+++ b/app/Commands/HelpCommand.php
@@ -14,6 +14,8 @@ class HelpCommand extends Command
         'enable {service}' => 'Enable the provided service',
         'disable' => 'Disable a service from a list of options',
         'disable {service}' => 'Disable the provided service',
+        'restart' => 'Restart a stopped container',
+        'restart {container}' => 'Restart a stopped container',
         'list' => 'List all enabled services',
     ];
 
--- a/app/Commands/RestartCommand.php
+++ b/app/Commands/RestartCommand.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace App\Commands;
+
+use App\InitializesCommands;
+use App\Services;
+use App\Shell\Docker;
+use Illuminate\Support\Str;
+use LaravelZero\Framework\Commands\Command;
+use PhpSchool\CliMenu\CliMenu;
+
+class RestartCommand extends Command
+{
+
+    use InitializesCommands;
+
+    protected $signature = 'restart {container?}';
+    protected $description = 'Restart a service.';
+
+    public function handle(): void
+    {
+        $this->initializeCommand();
+
+        $container = $this->argument('container');
+
+        if ($container) {
+            $this->restart($container);
+
+            return;
+        }
+
+        $this->menu('Containers to restart')
+            ->addItems($this->restartableContainers())
+            ->open();
+
+//        if (!$option) {
+//            return;
+//        }
+//
+//        $this->restart($option);
+    }
+
+    public function restartableContainers(): array
+    {
+        return collect(app(Docker::class)->takeoutContainers())->skip(1)->reduce(function ($containers, $container) {
+            if(!Str::contains($container[2], 'Up')) {
+                $containers->push(["$container[0] - $container[1]", function(CliMenu $menu) use ($container) {
+                    $this->restart($menu->getSelectedItem()->getText());
+
+                    foreach($menu->getItems() as $item) {
+                        if($item->getText() === "$container[0] - $container[1]") {
+                            $menu->removeItem($item);
+                        }
+                    }
+
+                    $menu->redraw();
+                }]);
+            }
+
+            return $containers;
+        }, collect())->toArray();
+    }
+
+    public function restart(string $container): void
+    {
+        if(Str::contains($container, ' -')) {
+            $container = Str::before($container, ' -');
+        }
+
+        app(Docker::class)->startContainer($container);
+    }
+}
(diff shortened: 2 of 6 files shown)


AlexMartinFR:
Good idea!

It would be worth offering all three `start`, `stop` & `restart` commands, in my opinion. 
I was expecting them and had to resort to native Docker commands instead 🙂

app/Commands/RestartCommand.php
```diff
@@ -0,0 +1,72 @@
...
+            $this->restart($container);
+
+            return;
+        }
+
+        $this->menu('Containers to restart')
+            ->addItems($this->restartableContainers())
+            ->open();
+
+//        if (!$option) {
```
mattstauffer, line 36: Any reason these lines are commented out?

app/Commands/StopCommand.php
```diff
@@ -0,0 +1,66 @@
...
+namespace App\Commands;
+
+use App\InitializesCommands;
+use App\Services;
+use App\Shell\Docker;
+use Illuminate\Support\Str;
+use LaravelZero\Framework\Commands\Command;
+use PhpSchool\CliMenu\CliMenu;
+
+class StopCommand extends Command
```
mattstauffer, line 12: Why do we want people to be able to stop services?

app/Commands/StopCommand.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $container = $this->argument('container');
+
+        if ($container) {
+            $this->stop($container);
+
+            return;
+        }
+
+        $this->menu('Containers to restart')
```
mattstauffer, line 32: Looks like some of the text here is out of date.

app/Commands/StopCommand.php
mattstauffer: Additionally, if this were to stay, you'd also need to add it to the notes in the help command.

AlexMartinFR: If I may, my use case for `stop`:

- I have an old MySQL container only needed when I work on legacy apps.
- A stopped container doesn't consume RAM nor CPU so I can keep it around for a while.
- It's easier to start a stopped container than to spin-up a new one and having to go through port, version & volume configuration again.

infernobass7: I agree with @AlexMartinFR. My personal laptop is a 2011 MacBook and if you mistype the volume name when reenabling the service would cause the data to not be loaded as expected.

jonsugar:
Could we change the command names? This could just be in my brain but...

To me `enable` means to start or switch on and `disable` means to stop or switch off. Take out thinks of them as to require or remove. Are we working more along the lines of composer?

```
takeout require mysql
takeout remove mysql
```

Requiring and removing sits better in my head since now I know to start/stop the containers in my docker dashboard or using the docker command line. Or as this PR introduces:

`takeout start|stop mysql` or `takeout start|stop TO--mysql--<tag>`

app/Commands/StopCommand.php
mattstauffer: This is a pretty compelling case. Thanks!

mattstauffer:
@jonsugar You're thinking  a little more like a Docker user here. We intentionally called them `enable` and `disable` to distance these commands from the idea of being like Composer, where you install it but it can be running or not running, because we couldn't think of a single compelling use case for why someone would need to *stop* a container, rather than just removing it. 

However.. in the comments a case has been made! Removing a container is just fine, *but* booting the same container with the exact same settings later is *harder* if you have to remove it instead of potentially stopping it... at which point it now makes sense again to consider differentiating  starting and stopping vs adding and removing.

So, *now* that we're doing that, there's a bit more a Composer-like-ness, *but* we're also starting it when we add it and stopping when we remove it.

This was what I was thinking about last week:

takeout add
takeout stop
takeout start
takeout remove

The biggest issue with those is that running them with no parameters feels a bit off--`takeout stop` sort of feels like it should stop takeout, but it's not going to. But I think I'm OK with that.

app/Commands/HelpCommand.php
```diff
@@ -14,6 +14,8 @@ class HelpCommand extends Command
         'enable {service}' => 'Enable the provided service',
         'disable' => 'Disable a service from a list of options',
         'disable {service}' => 'Disable the provided service',
+        'restart' => 'Restart a stopped container',
```
mattstauffer, line 17: Can you update this for `start` and `stop`?

mattstauffer:
All this said, if we limit this PR to `start` and `stop`, we can hopefully just merge it in and then consider renaming the other commands.

The biggest question here is, what's the most appropriate syntax to use for passing a command into `start` or `stop`? This PR shows container ID; Jon's suggest is the volume name.

I personally think the volume name is pretty clumsy so I'm inclined to go for container ID.

Any thoughts?

infernobass7:
I would use the container ID since that is displayed in the `takeout list` command. Optionally, leaving the container ID off of the command provides you with a list of takeout containers that are running or stopped for the user to chose from.

app/Commands/HelpCommand.php
infernobass7: Updated in most recent commit.

app/Commands/StartCommand.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $this->menu('Containers to start')
+            ->addItems($this->startableContainers())
+            ->open();
+    }
+
+    public function startableContainers(): array
+    {
+        return collect(app(Docker::class)->takeoutContainers())->skip(1)->reduce(function ($containers, $container) {
+            if(!Str::contains($container[2], 'Up')) {
```
mattstauffer, line 40: Can we move this `Str::contains` into a `reject` before we get into the reduce?

mattstauffer: Also... could this become a `mapWithKeys` instead of a `reduce`? What we're basically doing here is rejecting a bunch of containers then modifying the rest, right?

app/Commands/StopCommand.php
```diff
@@ -0,0 +1,66 @@
...
+use App\InitializesCommands;
+use App\Services;
+use App\Shell\Docker;
+use Illuminate\Support\Str;
+use LaravelZero\Framework\Commands\Command;
+use PhpSchool\CliMenu\CliMenu;
+
+class StopCommand extends Command
+{
+
```
mattstauffer, line 14: Drop this newline please

app/Commands/StopCommand.php
```diff
@@ -0,0 +1,66 @@
...
+    use InitializesCommands;
+
+    protected $signature = 'stop {container?}';
```
mattstauffer, line 17: OK for this to also be `containerId`?

app/Commands/StopCommand.php
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        $this->menu('Containers to stop')
+            ->addItems($this->stoppableContainers())
+            ->open();
+    }
+
+    public function stoppableContainers(): array
+    {
+        return collect(app(Docker::class)->takeoutContainers())->skip(1)->reduce(function ($containers, $container) {
```
mattstauffer, line 39: Same notes on this collection work as the other one

app/Commands/StartCommand.php
```diff
@@ -0,0 +1,66 @@
...
+use Illuminate\Support\Str;
+use LaravelZero\Framework\Commands\Command;
+use PhpSchool\CliMenu\CliMenu;
+
+class StartCommand extends Command
+{
+
+    use InitializesCommands;
+
+    protected $signature = 'start {container?}';
```
mattstauffer, line 17: Would it hurt at all to make this `containerId`

app/Commands/StartCommand.php
infernobass7: Is there any reason to use mapWithKeys instead of just map?

app/Commands/StartCommand.php
infernobass7: No. Makes more sense as `containerId`

app/Commands/StopCommand.php
infernobass7: Makes more sense as `containerId`

AlexMartinFR:
I'm all for this new command set 👍 

```
takeout add
takeout remove
takeout start
takeout stop
```

It makes much more sense than the actual, whether or not one is familiar with Docker!
That was the main pain point encountered during testing with a developer friend.

Thanks!

app/Commands/StartCommand.php
mattstauffer: I think I thought you were ending up with a key and a value; if `map` makes more sense then I probably just read it wrong!

mattstauffer:
Thanks!
I'm going to add a separate issue about writing tests for this; I want to merge it but I also want to get test coverage.

PR #72, opened 2020-08-27
feat: add MariaDB service

owenvoke:
This adds MariaDB as an alternative to MySQL.

- Website: https://mariadb.org
- Docker image: https://hub.docker.com/_/mariadb

~~I guess it may be worth thinking about https://github.com/tightenco/takeout/issues/53 before merging though, as I think this affects MariaDB as well.~~ 🤔 Looks like MariaDB advise against it though: https://mariadb.com/kb/en/authentication-plugin-mysql_native_password

Actually, it looks like this isn't necessary in PHP `~7.1.16 || >=7.2.4` as it was updated to support it and Takeout requires `^7.2.5`. 👍🏻

### DIFF
--- a/app/Services/MariaDB.php
+++ b/app/Services/MariaDB.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace App\Services;
+
+class MariaDB extends BaseService
+{
+    protected $imageName = 'mariadb';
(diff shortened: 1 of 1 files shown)


mattstauffer:
Thanks @owenvoke!!

PR #75, opened 2020-08-27
Added minio service

luisdalmolin:
Added Minio service (https://github.com/minio/minio) for a local "s3" instance.

### DIFF
--- a/app/Services/Minio.php
+++ b/app/Services/Minio.php
@@ -12,11 +12,21 @@ class Minio extends BaseService
             'prompt' => 'What is the Docker volume name?',
             'default' => 'minio_data',
         ],
+        [
+            'shortname' => 'access_key',
+            'prompt' => 'What will the access key for Minio be?',
+            'default' => 'minio',
+        ],
+        [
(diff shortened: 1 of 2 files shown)


app/Services/Minio.php
```diff
@@ -0,0 +1,22 @@
...
+    protected $prompts = [
+        [
+            'shortname' => 'volume',
+            'prompt' => 'What is the Docker volume name?',
+            'default' => 'minio_data',
+        ],
+    ];
+
+    protected $dockerRunTemplate = '-p "$port":9000 \
+        -e MINIO_ACCESS_KEY=minio
```
mattstauffer, line 18: I've never used Minio; is it likely people will want to change their access key and secret? If so, can you add those as prompts as well?

luisdalmolin: Yeah, I guess it makes sense to put as prompts with default values. Minio is pretty cool for having a local S3 for development.

luisdalmolin: Updated & pushed.

mattstauffer:
Thanks!!

PR #78, opened 2020-08-27
Update 🥡 README.md

AlexMartinFR:
- Adding a precision regarding the multiple MySQL containers example: they must not share the same volume
- Fixed PostgreSQL spelling


mattstauffer:
Great update! Thanks @AlexMartinFR!

AlexMartinFR:
Thanks 🙂

PR #82, opened 2020-08-28
Add DynamoDB service

cappuc:
This PR had service for DynamoDB.
I used [cnadiminti/dynamodb-local](https://hub.docker.com/r/cnadiminti/dynamodb-local) docker image instead of official one from amazon [amazon/dynamodb-local](https://hub.docker.com/r/amazon/dynamodb-local) because the amazon image uses in memory db and is not configurable. In order to use a volume for the database with official image we should override the image command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ But you can also easily enable ElasticSearch, PostgreSQL, MSSQL, Redis, and more
 - Memcached
 - MailHog
 - MariaDB
+- DynamoDB
 
 ## Requirements
 
--- a/app/Services/DynamoDB.php
+++ b/app/Services/DynamoDB.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace App\Services;
+
+class DynamoDB extends BaseService
+{
+    protected $organization = 'cnadiminti';
+    protected $imageName = 'dynamodb-local';
+    protected $defaultPort = 8000;
+    protected $prompts = [
+        [
+            'shortname' => 'volume',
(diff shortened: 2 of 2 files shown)


mattstauffer:
Thanks @cappuc! Are there any downsides to using this unofficial version? Since I'm unfamiliar with it, I just want to make sure I don't commit to this and unknowingly introduce some issues. Thanks!

cappuc:
Looking at the [Dockerfile](https://github.com/cnadiminti/docker-dynamodb-local/blob/master/Dockerfile) for cnadiminti/dynamodb-local on github, it simply download the dynamo db local executable and setup volume and init command.
The only downside I can see is that it can be outdated compared to the official one but I don't think there are major changes on dynamodb frequently.

I tried to make the official image working with persistent database and volumes but I failed because volume is mounted as root and the image uses a different user (dynamodblocal) so it doesn't have permissions to write to the volume

cappuc:
Hi, today I used cnadiminti/dynamodb-local image for a project and I noticed that the latest tag is not updated to the latest version of dynamodb on docker hub so it's better to find another solution.

cappuc:
I managed to get the official amazon image work with persistent database changing the container user to root

josecanhelp:
Nice job! This looks great.

I was able to run the script [here](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GettingStarted.NodeJs.01.html) against my local dynamoDB instance.

image

All checks are passing now, so I'll get this merged! 💪

PR #84, opened 2020-08-29
Update README.md to add an FAQ item about Sequel Pro and MySQL 8

ian-nisbet:
I had trouble with SequelPro when I first started using Takeout.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,12 @@ The solution is to just close your database GUI, and then it should be released.
 Why would you use this instead of compiling your dependencies from scratch? Better yet, why not write your own database engine? It's the only way to know there's not an NSA backdoor in it.
 </details>
 
+<details>
+<summary><strong>Sometimes I cannot connect to MySql with SequelPro at 127.0.0.1</strong></summary>
+
+You can try 0.0.0.0 in this case. Make sure that the port is correct.
+</details>
+
 ## Future plans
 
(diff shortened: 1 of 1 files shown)


josecanhelp:
@ian-nisbet I am able to replicate connection issues with sequel pro when using mysql 8. But mysql 5.7 works fine.

When attempting to connect with sequel pro using mysql 8, using `0.0.0.0` did not help. This is likely an issue with how sequel pro is connecting to mysql 8 and it isn't really something we can fix on our end.

Can you help me verify that this is the case? If so, maybe we can add a warning in the README about this specific situation.

mattstauffer:
@josecanhelp OK, so @ian-nisbet is saying switching to `0.0.0.0` helped but you're saying it didn't? Just want to make sure--thanks!

josecanhelp:
@mattstauffer I just attempted to reproduce this again. I was able to connect with Sequel Pro using `127.0.0.1` without any issue. Unfortunately, Sequel Pro hasn't had a release since 2016 so it is known to be finicky.

image

josecanhelp:
@ian-nisbet I am going to close this ticket out. Sequel Pro should likely be avoided as a client. Per Freek, there is a more updated version called [Sequel Ace](https://sequel-ace.com/).

https://twitter.com/freekmurze/status/1301030834879770625

PR #87, opened 2020-08-30
Update port warnings to better handle some edge cases

josecanhelp:
## Problem
Currently, if the connected database client is left open and a container is disabled, then the port is "read' as being in use.

## Suggested Solution
Changing from `netstat` to `lsof`, we can filter the results of port usage to only those being used by docker.

## New expectation
This should allow the port usage warning to be more accurate during `takeout enable`.


mattstauffer:
@josecanhelp Do we lose anything here because a port might actually be legitimately occupied by a non-Docker device? For example, when I couldn’t install Mailhog because the port was already taken, this would’ve kept me from noticing it, and I assume I still would’ve run into an issue, right?

josecanhelp:
Ah yea, you're right. This would limit the warning to docker based usage. Another option would be to add a "Check open clients" to the warning.

michaeldyrynda:
I think the reason that this is biting me on the bum is because I have to ssh tunnel through some jumpboxes to get to my database servers at work.

As a result, the netstat command will return a false positive:

```
netstat -vanp tcp | grep 3306
tcp4       0      0  127.0.0.1.33306        *.*                    LISTEN      131072 131072  15081      0 0x0100 0x00000006
tcp6       0      0  ::1.33306              *.*                    LISTEN      131072 131072  15081      0 0x0100 0x00000006
```

Meanwhile, `lsof` doesn't have this problem, though it likely still leaves the issue @mattstauffer suggested.

I think that using a regular expression ought to be enough to avert my issue:

```
netstat -vanp tcp | grep "\.3306"
```

Therefore, some variant of the following ought to do the trick:

```php
$process = $this->shell->execQuietly(sprintf('netstat -vanp tcp | grep "\.%s"', $port));
```

mattstauffer:
Could we switch to `lsof` but *not* limit it to ports taken by Docker? or would Michael's most recent example be better?

michaeldyrynda:
I think that sticking with `netstat` is probably the better option as it's a more complete check of open connections, rather than open file handles. The `grep` as a regex should give the necessary specificity to avoid edge cases as above.

josecanhelp:
@mattstauffer I agree with @michaeldyrynda. We should stick with `netstat` and use a regular expression. We will still need to add a message for users to check their database clients. 

We could technically regex filter results with `CLOSE_WAIT` or `FIN_WAIT_1`, which is what a non-closed client status will look like. But that might be getting too clever.

michaeldyrynda:
One other downside to this approach is if I connect to say, a public RDS instance, it'll still come up in the `netstat` output.

```
tcp4       0      0  10.1.1.107.57243       xxx.xxx.xxx.xxx.3306   ESTABLISHED   131072 131768  75881      0 0x2131 0x00000008
```

It'll also appear for a period after closing the connection, too as `TIME_WAIT`.

josecanhelp:
I don't see any issue for filtering out results with `TIME_WAIT`. We really just want to make sure any new services won't be overriding currently running ports.

danielebarbaro:
Hi, i open a new PR #95 maybe we can use another method to check who and what use the selected port.

mattstauffer:
@danielebarbaro Thanks! Since I don't really understand this stuff, I'll have to ask: does your PR solve the reasons why @michaeldyrynda and @josecanhelp suggested netstat is better?

Thanks!!

danielebarbaro:
Hi, @mattstauffer  and thanks for your reply :)

I misunderstood the "issue" posted by @josecanhelp.

I personally use Brew and Takeout and it happens that something goes wrong because I forget to turn off the Brew service.

I proposed PR #95 to understand and clarify which process, PID and user still use that port (docker or brew).

If we don't care who and what are using that port, the usage of nestat (show network status) with the @michaeldyrynda regex is more correct than lsof (list open files).

josecanhelp:
@mattstauffer can I get a review on this when you get a chance? I think the new command will work much better.

cc: @michaeldyrynda thanks for the help on this one!

thecrypticace:
I'd like to see this in too. I'm getting a false positive b/c 'grep 3306' also picks up port '33060'. I had to stop homestead temporarily to enable mysql.

mattstauffer:
@josecanhelp Reviewed. Big important fix here, thanks!! Thanks @michaeldyrynda!

Issue #77, opened 2020-08-27
Include the volume name in the container name to prevent duplicates 


dducro:
First of all, thanks for creating this easy tool!
This already was exactly my work flow, creating docker containers for just the services like ElasticSearch, Redis etc. This makes maintaining them a lot easier! :D

I'd like to create a separate Elastic search and Redis container for each project.
However, if they both use the same version there is a name conflict.

This could be solved by including the volume name in the container name like: `TO--redis--volume-name--buster`

What are your thoughts on this?
Thanks!

```
ERR  docker: Error response from daemon: Conflict. The container name "/TO--redis--buster" is already in use by container "0d0d690f64328b59d4a2b60f54cd99ee2254ba48f63db302b1738ca927347410". You have to remove (or rename) that container to be able to reuse that name.
See 'docker run --help'.
```

mattstauffer:
I think this is a great idea! I'm trying to decide whether it's a breaking change or not. I.. sort of think it isn't? So.. we may be able to add this in a dot release.

dducro:
Thanks for your answer! :)
I see that not all services have a volume, like: MsSql, Memcached, MailHog.
This means the container name can't always include the volume in its name.
In these cases it would still be nice to optionally be able to provide a name to make the containers unique.
I don't think it's a breaking change because all containers starting with `TO--` are listed an all commands are done with its ID.
Please let me know if you would like me to work on a PR.

(Just curious, doesn't MsSql need a volume as well?)

josecanhelp:
@dducro MsSql could benefit from a volume, but the official image has trouble mounting docker volumes for some reason. In order to still have the service available in Takeout, we needed to disable volumes for that service.

I have seen some tricks to make volumes work with MsSql, but for now it's not available. So you can bring up the container, but as soon as you destroy it, the data will be removed.

We are looking into the option to bring down a container without destroying it though, so that will make working with MsSql a little better.

mattstauffer:
The only way I can think of so far is to add an additional prompt to the setup that's basically the "nickname" or something, which would default to the volume name, and the nickname would always be appended to the base name (e.g. `TO--mysql--5.7--nicknamehere`)

dducro:
That would be perfect! This way the user could have an instance per project if he so desires.
I also use the docker desktop app to list the containers and turn them on/off, to see the project name in the container name would be a great improvement! :)

mattstauffer:
@dducro OK, it's a plan!

nehero:
[...]

PR #89, opened 2020-08-31
Specify an optional nickname when enabling a service

ozziexsh:
This PR is based on the discussion happening in #77 

## The Problem

When creating multiple service containers of the same version you will run into naming conflicts. 

ex:

```
 ERR  docker: Error response from daemon: Conflict. The container name "/TO--mariadb--focal" is already in use by container "1605cd2ae8729609ec3e62e43268e92ec2891c27f9d7ae9fcf2bb8c302c39718". You have to remove (or rename) that container to be able to reuse that name.
See 'docker run --help'.
```

## The Solution

Prompt the user for a nickname (defaults to a `uniqid()`) that we add to the container name to ensure uniqueness

## Considerations

* No tests yet
* I chose the word `nickname` based on the discussion in the issue
* We could do a retry prompt similar to the one for choosing a port if there is a name conflict
* I chose to use a unique string instead of the volume name as a fallback, but we could easily add that in

Happy to hear feedback 😄

### DIFF
--- a/app/Services/BaseService.php
+++ b/app/Services/BaseService.php
@@ -31,6 +31,11 @@ abstract class BaseService
             'prompt' => 'Which tag (version) of this service would you like to use?',
             'default' => 'latest',
         ],
+        [
+            'shortname' => 'nickname',
+            'prompt' => 'Enter a nickname for this container',
+            // Default is set in the constructor
+        ],
     ];
     protected $prompts;
     protected $promptResponses = [];
@@ -48,6 +53,9 @@ abstract class BaseService
         $this->defaultPrompts = array_map(function ($prompt) {
             if ($prompt['shortname'] === 'port') {
                 $prompt['default'] = $this->defaultPort;
+            }
+            if ($prompt['shortname'] === 'nickname') {
+                $prompt['default'] = uniqid();
             }
             return $prompt;
         }, $this->defaultPrompts);
@@ -95,6 +103,11 @@ abstract class BaseService
     public function shortName(): string
     {
         return strtolower(class_basename(static::class));
+    }
+
+    public function nickname(): string
+    {
+        return Str::slug($this->promptResponses['nickname']);
     }
 
     public function defaultPort(): int
@@ -155,6 +168,6 @@ abstract class BaseService
 
     protected function containerName(): string
     {
-        return 'TO--' . $this->shortName() . '--' . $this->tag;
+        return 'TO--' . $this->shortName() . '--' . $this->nickname() . '--' . $this->tag;
     }
 }
--- a/app/Services/BaseService.php
+++ b/app/Services/BaseService.php
@@ -107,7 +107,7 @@ abstract class BaseService
 
     public function nickname(): string
     {
-        return Str::slug($this->promptResponses['nickname']);
+        return Str::slug($this->promptResponses['nickname'] ?? '');
     }
 
     public function defaultPort(): int
@@ -168,6 +168,13 @@ abstract class BaseService
 
     protected function containerName(): string
     {
-        return 'TO--' . $this->shortName() . '--' . $this->nickname() . '--' . $this->tag;
+        return collect([
+            'TO',
+            $this->shortName(),
+            $this->tag,
+            $this->nickname(),
+        ])->filter(function($str) {
+            return !empty($str);
+        })->join('--');
     }
 }


app/Services/BaseService.php
```diff
@@ -155,6 +168,6 @@ protected function buildParameters(): array
 
     protected function containerName(): string
     {
-        return 'TO--' . $this->shortName() . '--' . $this->tag;
+        return 'TO--' . $this->shortName() . '--' . $this->nickname() . '--' . $this->tag;
```
mattstauffer, line 171: Can we put the nickname at the end and only add the `--` if there's a nickname?

nehero: For sure! Remove the default unique string too?

mattstauffer: @nehero Oh, I hadn't even noticed that, actually!

I think... Let me think about it a bit. :)

michaeldyrynda:
Is there a limit to the length of the container names? Being mindful of possibility of truncation.

The image includes the tag name i.e. `redis:buster` or `mariadb:10.3.22`, so probably not necessary to append it to the container name as well.

Doing it this way means you can use `$this->nickname()`, with a default `uniqid()` if not specified.

nehero:
> Is there a limit to the length of the container names? Being mindful of possibility of truncation.

Just tried creating one with 500+ characters and no issues. Couldn't find a reliable source on any hard cap for container names on a local docker instance. 

IMO if there's no text cap the extra information in the name is a nice convenience but yeah having the uniqid fallback in there is nice if you're just entering through defaults you won't run into the name conflict

mattstauffer:
@nehero I think I like it! I'm sorry I made you do the work you did in your latest commit; you're right that nickname makes the most sense with the `uniqid` fallback, which means you don't have to worry about an empty one.

I *do* still like the nickname at the end, though.

@josecanhelp mentioned something about appending the nickname to the volume name as well, but I don't totally follow the workflow for that--Jose, would you mind explaining how that would work? 

Thanks!

nehero:
@mattstauffer Haha all good!

Yeah having the nickname in the volume is ideal so multiple containers aren't sharing a filesystem

@josecanhelp a quick solution we could add is

```php
if ($prompt['shortname'] === 'volume') {
  $this->promptResponses[$prompt['shortname']] .= '_' . $this->promptResponses[$prompt['nickname']];
}
```

And would it be more efficient to just move the volume prompt to the `defaultPrompts` and allow specific services that don't utilize it to opt-out? We could allow them to specify their default value like `mariadb_data` through a class property like `protected $defaultVolume = ""`

Let me know your thoughts and I can make the changes

josecanhelp:
So if we append the nickname to the volume BUT ALSO allow the users to set the default volume name, it would be a little confusing to the user when they'd want to bring up a container with that same volume because they didn't append the nickname the first time they named the volume.

I'm almost thinking we should make all of the volume names just $serviceName_some_hash_$nickName. Thoughts on this @mattstauffer or @nehero?

Or better yet...

### During Service Enable (for example mysql)
#### If the following are true:
- A volume is prepended with `mysql_` [docker volume ls]
- That volume is not already mounted in an another container [docker inspect all running takeout containers]
#### Then
- Allow the user to create a new volume or (still not sure exactly how to do this one)
- Allow the user to select one of the `mysql_` volumes from a menu

mattstauffer:
Hm. Yah, appending anything to the volume makes attaching to that volume again tough.

What if, instead of checking for `mysql_`, we just allow them to attach to *any* existing unused volume? A bit more flexibility that way, so they could attach Takeout-spun-up services to existing volumes.

nehero:
If we did that then I'd be curious how the flow would work, going from either specifying a volume name to choosing one from a list. Unless the volume is the last step where it shows a screen similar to where you pick the service:

```
What volume should we use for this?
- [ ] mysql_data
- [ ] my_other_volume
- [ ] create me a new one
```

And then creating them a new one would do like what jose said with the convention `$serviceName_some_hash_$nickName`

mattstauffer:
I wonder if we did it a bit like that:

```
What volume should we use for this?
DEFAULT VOLUME NAME
- [ ] mysql_data

EXISTING UN-ATTACHED VOLUMES
- [ ] my_other_volume

- [ ] Let me specify a new one
```

And then if they chose the last one they could type whatever they want

nehero:
We could for sure make that work. Let me know if you'd like it in this PR or a separate one

josecanhelp:
@mattstauffer I realize there was a "solution" overlap when I merged in some work that appended the used port to the container name. This did technically solve the issue with multiple containers of the same service being enabled and their names won't conflict. It is more automated than asking the user for a nickname, but I do like the idea of adding a nickname if users want to tag containers for specific projects or something.

Should we roll back the concatenation of the port number to the container name and move forward with nicknames, or should we hold off on the nickname feature? Or keep both?

mattstauffer:
@josecanhelp Honestly I'm not sure! What if.. we allowed for a nickname but it used the port by default, so we get best of both worlds? Customizability together with a good default?

josecanhelp:
@mattstauffer I like that. I'm just weary of adding more steps to the enable prompt. The nickname feature should provide enough value to make it worth adding another prompt. Do you think it does this?

mattstauffer:
@josecanhelp ahhhghghghg i don't know it's so hard lol.. especially because a lot of folks who *do* need this differentiation would likely consider doing docker compose for dependency per project.

However.. I *could* imagine needing a nickname if I have multiple versions of MySQL, for example, and I need to attach to the container that works for *this* version.. which I don't think our current system delineates, right? So ... maybe it is worth it.

jonsugar:
Maybe I've missed something but can we just release a version 2.0 (breaking changes) where the user MUST nickname new containers. Takeout would create the volume in the background using the nickname. The user would have no say in the volume name. This way users can have as many of the same version (label) of a container as they want. All the problems go away.

I'm not sure there is a use case where users need to attach a container to an existing volume.

mattstauffer:
Reviewing this again and I have seen a lot of noise from folks lately wanting the things to spin up *faster*, not *slower*.

Is this problem primarily solved by appending the port to the container name? If so, can we call this done for now?

ozziexsh:
@mattstauffer It'd be easy enough to remove all of the nickname stuff here and just append the port at the end and honestly since the ports have to be unique it'd solve the problem of name clashes. Agreed that adding the extra nickname prompt could get in the way but would be easy to reintroduce down the line if needed

josecanhelp:
I think adding the port number to the container name has resolved MOST conflict issues. Adding nickname might make it too confusing. Thanks for the input!

PR #93, opened 2020-09-02
Added in ability to detect OS and return correct command for Docker

Ravelin:
Added in ability to check OS and return the correct command for different operating systems.

Windows defaulted to powershell

Screenshot from 2020-09-02 10-59-09

### DIFF
--- a/app/Shell/Docker.php
+++ b/app/Shell/Docker.php
@@ -8,10 +8,29 @@ use Symfony\Component\Process\Process;
 class Docker
 {
     protected $shell;
+    protected $cmd;
 
     public function __construct(Shell $shell)
     {
         $this->shell = $shell;
+        $this->cmd = $this->detectDockerCommand();
+    }
+    
+    //Checks for OS version
+    // returns command for OS
+    private function detectDockerCommand(): string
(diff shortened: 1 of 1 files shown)


josecanhelp:
Thanks for this work! I was able to get the menu to display in Linux, but all of the other commands fail to run and I get this message. Are you able to run containers in linux with your changes here?

Screen Shot 2020-09-04 at 1 57 13 AM

Ravelin:
I was able to get containers to run on my system, however looking at the code now it would probably be better to use a $which docker and it'll find and run the docker container. I'll do the changes and re-push if that is ok

Ravelin:
Just double checked and I am getting a docker container pull and install.  I've updated my local code to 
`return 'pgrep -f  $which docker';`

in order to counter if docker is installed in a none standard location

Screenshot from 2020-09-04 20-22-01

[screenrecording.zip](https://github.com/tightenco/takeout/files/5176657/screenrecording.zip)

Also have you allowed docker to run as a non-root user inside your VM? https://docs.docker.com/engine/install/linux-postinstall/ ?

Issue #58, opened 2020-08-25
Consider adding a `ports` debugger command


mattstauffer:
If someone is surprised by the ports taken up on their machine, we may be able to simplify discovering what's running there.

josecanhelp:
This is outside the scope of takeout. We just need to check if the port is in use, which has since been included.

PR #95, opened 2020-09-03
Use lsof to better handle port info

danielebarbaro:
Hi, this pull req solve in another way this problem: #87 (issue #58 )

## Problem
Currently, if the connected database client is left open and a container is disabled, then the port is "read' as being in use.

## Suggested Solution
Add a new method (`portInfo($port)`) to render process info with lsof.
`lsof -n -P | grep -Ei 'ipv4.*:{$port}' |  tr -s ' ' | cut -d' ' -f1,2,3`

## New expectation
Inform which pid, process and user still using that port.


josecanhelp:
Closing per our conversation in #87 . We'll use netstat to check the ports. Thanks @danielebarbaro!

PR #99, opened 2020-09-04
Add beanstalkd

mattstauffer:
Also allow for images with only "latest" tag, and services with no prompts


josecanhelp (approved):
Nicely done!

PR #103, opened 2020-09-09
Fixing minio run template bug '\'

khanalpride:
Just added the \ to fix the minio run command.


mattstauffer:
Thanks @khanalpride!

Issue #107, opened 2020-09-10
MongoDB Support


bsormagec:
Hello friends,

Can you add MongoDB Support please.

PR #108, opened 2020-09-11
Add Mongo support

danielebarbaro:
This PR solve #107


josecanhelp (approved):
Excellent! I was able to connect to the mongo service via TablePlus using the url string:
`mongodb://admin:password@localhost:27017/admin`

LGTM!

PR #116, opened 2020-09-12
Fix MySQL default auth for PHP < 7.4

### LINKED ISSUE #53: MySQL 8: Change default auth on bootup via docker command
MySQL 8 requires (still, I believe!) the use of the `--default-authentication-plugin=mysql_native_password` flag to work with Laravel (and really just stay inline with user expectations).

> @josecanhelp - let me know if I'm wrong on that, I still use 5.7 for the most part!

One way of doing that would be to edit the default service templates to add a `$cmd` variable, for example, MySQL's:

```php
protected $dockerRunTemplate = '-p "$port":3306 \
        -e MYSQL_ROOT_PASSWORD="$root_password" \
        -v "$volume":/var/lib/mysql \
        "$organization"/"$image_name":"$tag" "$cmd"';
```

Then we'd have to handle allowing a custom `$cmd` to be used when starting a container. 

> For MySQL, the `ENTRYPOINT` is (for all intents and purposes) `mysqld`. The Docker `CMD` can just be flags to pass the `mysqld` command, like so:
> 
> ```bash
> docker run -d mysql:8 --default-authentication-plugin=mysql_native_password
> ```

Adding a `$cmd` to the templates would allow users to add that flag if they wanted.

This addition may mean another default prompt for services, but perhaps it could be a hidden option for advanced usage?

Another idea in this case is to code up a ✨special case conditional✨😅 where MySQL 8+ gets this flag appended to `$cmd`.

thecrypticace:
The alternative is to edit the user after the container is created but that’s less than ideal. This should “just work” out of the box when the user is using an older PHP version.

I opted to not do detection because it's possible that the PHP version connecting to MySQL is different than the one used to start the container (e.g. if you're using takeout on the host machine + an older homestead box)

fixes #53

### DIFF
--- a/app/Services/MySql.php
+++ b/app/Services/MySql.php
@@ -22,7 +22,7 @@ class MySql extends BaseService
     protected $dockerRunTemplate = '-p "${:port}":3306 \
         -e MYSQL_ROOT_PASSWORD="${:root_password}" \
         -v "${:volume}":/var/lib/mysql \
-        "${:organization}"/"${:image_name}":"${:tag}"';
+        "${:organization}"/"${:image_name}":"${:tag}" --default-authentication-plugin=mysql_native_password';
 
     protected static $displayName = 'MySQL';
 }


mattstauffer:
Merging, going to test on a few versions to attempt to see if it has any issues. Thanks @fideloper and @thecrypticace!

Issue #76, opened 2020-08-27
Use the official elastic registry for elasticsearch


abenettt:
Elastic hosts their docker images on `docker.elastic.co`, we should consider pointing there when looking for the service.

PR #119, opened 2020-09-14
Add docker.elastic.co support for elasticsearch images

danielebarbaro:
Hi! 
This PR solve #76 .

Official **elasticsearch** images are in _https://www.docker.elastic.co/r/elasticsearch_.
The problem was that _https://docker.elastic.co/v2/elasticsearch/elasticsearch/tags/list needs_ an `access_token`.

I wrote some lines of code to catch a valid `access_token` before call tag list url.


abenettt:
This looks great!
Do you know how long the access token is valid for? Maybe if it is for long enough we could cache it for a time so we don't have to request it always.

danielebarbaro:
Hi @abenettt, unfortunately there's no `expire_in` field on auth response.

I've been thinking about saving the token somewhere but you need to take a valid token only when you create a new container.

abenettt:
Ah yeah you're right it would not really make sense to save the token.

danielebarbaro:
hi, @josecanhelp @mattstauffer I just updated the PR just to keep branch cleaned. Feel free to discard this PR

josecanhelp:
Great work, @danielebarbaro. Thanks!

PR #127, opened 2020-09-17
Allow enabling/disabling multiple services at once

bakerkretzmar:
Something @loganhenson said got me thinking about making it even smoother to spin multiple services up at once, so just PRing this as an idea. Before:

```bash
takeout enable meilisearch
# ...
takeout enable beanstalkd
```

After:

```bash
takeout enable meilisearch beanstalkd
```

I installed docker and manually tested this a bit and it works great, but I only really explored the happy path. I'm pretty sure this would still play nicely with other options or flags being passed to `enable` if that were added at some point.

### DIFF
--- a/app/Commands/DisableCommand.php
+++ b/app/Commands/DisableCommand.php
@@ -11,7 +11,7 @@ class DisableCommand extends Command
 {
     use InitializesCommands;
 
-    protected $signature = 'disable {serviceNames*}';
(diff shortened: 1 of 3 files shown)


mattstauffer:
Looks great to me! @josecanhelp any thoughts?

josecanhelp (approved):
THANKS!!

Issue #123, opened 2020-09-17
Don't allow for attaching a new container to a volume that's already in use by another running container


mattstauffer:

danielebarbaro:
Hi, with this issue you want to know:
 - if a volume exist a volume exist and

PR #128, opened 2020-09-18
Return right container name when a docker run fail.

danielebarbaro:
Hi @mattstauffer, i found and fix this:
Screenshot 2020-09-18 at 12 31 12

### DIFF
--- a/app/Shell/Docker.php
+++ b/app/Shell/Docker.php
@@ -120,7 +120,9 @@ class Docker
         $process = $this->shell->exec('docker run -d --name "${:container_name}" ' . $dockerRunTemplate, $parameters);
 
         if (! $process->isSuccessful()) {
-            throw new Exception("Failed installing {$containerName}");
+            $containerName = array_key_exists('container_name', $parameters) ? $parameters['container_name'] : '';
+            $imageName = $parameters['image_name'];
+            throw new Exception("Failed installing {$imageName}. The container {$containerName} is already in use.");
         }
     }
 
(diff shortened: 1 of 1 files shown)


app/Shell/Docker.php
```diff
@@ -120,7 +120,9 @@ public function bootContainer(string $dockerRunTemplate, array $parameters): voi
         $process = $this->shell->exec('docker run -d --name "${:container_name}" ' . $dockerRunTemplate, $parameters);
 
         if (! $process->isSuccessful()) {
-            throw new Exception("Failed installing {$containerName}");
+            $containerName = array_key_exists('container_name', $parameters) ? $parameters['container_name'] : '';
+            $imageName = $parameters['image_name'];
+            throw new Exception("Failed installing {$imageName}. The container {$containerName} is already in use.");
```
mattstauffer, line 125: Is this the only reason we can have this failure? Feels like all we can do right now is just throw the error but with your corrected way of pulling the container name

danielebarbaro: yes, the only reason is when a docker run fail. We can't catch the right container name.

The Docker class does not extend BaseService so we can't get containerName. 
I found this little "bug" when i wrote #123

mattstauffer: Right, but it's _possible_ the Docker run could fail for other reasons, right? Let me suggest an alternative.

mattstauffer, line 125:
```suggestion
            throw new Exception("Failed installing {$parameters['imageName']}.");
```

danielebarbaro: Yes, it is possible.
I changed the exception message to include the image name and a generic sentence about the container name because I don't know/i'm not sure how docker can fail.

danielebarbaro: It's more generic. I can change my PR if you want 😃

mattstauffer:
Thanks @danielebarbaro!

PR #129, opened 2020-09-18
Verify if a volume is active.

danielebarbaro:
Hi @mattstauffer , this maybe solve #123 .

I used "maybe" because I've a doubt. 
At the moment I checked with `docker ps -a --filter volume=volume_name` that a volume exist and the result is:
Screenshot 2020-09-18 at 12 37 55

So when user chose an existing container takeout return an error until user chose a unique name container.

Do you want to handle ,in some way, also this behaviour?

Screenshot 2020-09-18 at 12 38 09

### DIFF
--- a/app/Shell/Docker.php
+++ b/app/Shell/Docker.php
@@ -66,11 +66,6 @@ class Docker
         return $this->containerRawOutputToCollection($this->isVolumeUsedRawOutput($volumeName))->count() === 0;
     }
 
-    public function usedVolumes(string $volumeName): Collection
-    {
-        return $this->containerRawOutputToCollection($this->isVolumeUsedRawOutput($volumeName));
-    }
-
     /**
      * Given the raw string of output from Docker, return a collection of
      * associative arrays, with the keys lowercased and slugged using underscores
--- a/app/Shell/Docker.php
+++ b/app/Shell/Docker.php
@@ -63,7 +63,7 @@ class Docker
 
     public function volumeIsAvailable(string $volumeName): bool
     {
-        return $this->containerRawOutputToCollection($this->isVolumeUsedRawOutput($volumeName))->count() === 0;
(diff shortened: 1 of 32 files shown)


mattstauffer:
I think the second screenshot's behavior is the target of #89, so we can ignore that for this PR. Reviewing the rest of the PR now. Thanks!

app/Shell/Docker.php
```diff
@@ -61,6 +61,16 @@ public function allContainers(): Collection
         return $this->containerRawOutputToCollection($this->allContainersRawOutput());
     }
 
+    public function volumeIsAvailable(string $volumeName): bool
+    {
+        return $this->containerRawOutputToCollection($this->isVolumeUsedRawOutput($volumeName))->count() === 0;
+    }
+
+    public function usedVolumes(string $volumeName): Collection
```
mattstauffer, line 69: We can drop this, since it's not used, right?

app/Shell/Docker.php
```diff
@@ -93,6 +103,12 @@ protected function allContainersRawOutput(): string
         return trim($this->shell->execQuietly($dockerProcessStatusString)->getOutput());
     }
 
+    protected function isVolumeUsedRawOutput(string $volumeName): string
```
mattstauffer, line 106: What if we rename this `listMatchingVolumesRawOutput`?

mattstauffer (changes requested):
Great work! Just a few quick notes

danielebarbaro:
Hi @mattstauffer , thanks for your review. A made some review and a merge main in my branch to avoid conflicts.

danielebarbaro:
Hi @mattstauffer now PR is updated.
All tests pass.

This PR needs #157 .

mattstauffer:
@danielebarbaro Great--taking a look now. Thanks!

mattstauffer:
Looks great. Thanks so much @danielebarbaro!

Issue #88, opened 2020-08-31
Group the services list by categories


AlexMartinFR:
## Issue:

Takeout currently sorts the services list by alphabetical order.
As more and more services will be added to the list, it might become a bit confusing to browse.

## Suggestion:

Each service could have a `category` property, which could be used to at least group similar services.

Examples of categories:
- database
- storage
- search
- cache
- mail

## Improvement:

The`takeout enable` command could now have two steps instead of one:

- Select a category of services
- Choose the relevant service in the curated list

mattstauffer:
I don't know if I want extra steps, but I do like the idea of grouping.

I wonder if there's a way we could build a faster UI or a UI with more information available without adding an extra step?

AlexMartinFR:
I think [cli-menu's separators](https://github.com/php-school/cli-menu/blob/master/README.md) could be a good start:

![image](https://user-images.githubusercontent.com/19224681/91741874-9fc02880-ebb5-11ea-8b00-b4abbca68c5d.png)


Extra steps would only be necessary when the service list begins to be too long. The same way multi-steps forms are only relevant in some cases.

mattstauffer:
Yah, I like that!

AlexMartinFR:
Bear with me here, as this would need serious refactoring and is for now only generating the menu, but here is the raw idea:

```php
public function handle(): void
{
    […]

    $databaseServices = [
        'mariadb' => 'MariaDB',
        'mysql' => 'MySQL',
        'mssql' => 'MS SQL'
    ];

    $searchServices = [
        'meilisearch' => 'MeiliSearch',
        'elasticsearch' => 'ElasticSearch'
    ];

    $option = $this->menu('Services to enable')->setTitleSeparator('=');

    /* displayCategory() private method */
    $option->addStaticItem('Database:')->addStaticItem('---------');

    foreach ($databaseServices as $service => $name) {
        $option->addOption($service, $name);
    }

    $option->addLineBreak('', 1);
    /* End of the displayCategory() private method */

    /* Start of the displayCategory() private method */
    $option->addStaticItem('Search:')->addStaticItem('-------');

    foreach ($searchServices as $service => $name) {
        $option->addOption($service, $name);
    }

    $option->addLineBreak('', 1);
    /* End of the displayCategory() private method */

    $option->setPadding(2, 5)->open();

    if (!$option) {
        return;
    }

    $this->enable($option);
}
```

## Terminal preview:

[...]

PR #135, opened 2020-09-24
Group the services list by categories

danielebarbaro:
This PR solve #88 88

### DIFF
--- a/app/Commands/EnableCommand.php
+++ b/app/Commands/EnableCommand.php
@@ -4,7 +4,6 @@ namespace App\Commands;
 
 use App\InitializesCommands;
 use App\Services;
-use Illuminate\Support\Str;
 use LaravelZero\Framework\Commands\Command;
 
 class EnableCommand extends Command
@@ -25,16 +24,24 @@ class EnableCommand extends Command
             return;
         }
 
-        $option = $this->menu('Services to enable', $this->enableableServices())
-            ->addLineBreak('', 1)
-            ->setPadding(2, 5)
-            ->open();
+        $option = $this->menu('Services to enable:')->setTitleSeparator('=');
 
-        if (! $option) {
+        foreach ($this->enableableServicesByCategory() as $category => $services) {
+            $menuItems = collect($services)->mapWithKeys(function ($service) {
+                return [$service['shortName'] => $service['name']];
+            })->toArray();
+
+            $option->addStaticItem("{$category}:")
+                ->addStaticItem('---------')
+                ->addOptions($menuItems)
+                ->addLineBreak('', 1);
+        }
+
+        if (!$option) {
             return;
         }
 
-        $this->enable($option);
+        $this->enable($option->open());
     }
 
     public function enableableServices(): array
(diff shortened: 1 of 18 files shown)


app/Commands/EnableCommand.php
```diff
@@ -25,10 +24,20 @@ public function handle(): void
...
-            ->open();
+        $option = $this->menu('Services to enable:')->setTitleSeparator('=');
+
+        foreach ($this->enableableServicesByCategory() as $category => $services) {
+            $menuItems = collect($services)->mapWithKeys(function ($service) {
+                return [$service['shortName'] => $service['name']];
+            })->toArray();
+
+            $option->addStaticItem("{$category}:")
+                ->addStaticItem('---------')
```
AlexMartinFR, line 35: Nice! Quick visual refinement suggestion:

Custom separator length for each category to Tighten (pun intended) everything up!

```php
$separatorLength = strlen($category) + 1;
$separator = str_repeat('-', $separatorLength);

$option->addStaticItem("{$category}:")
    ->addStaticItem($separator)
    […]
```

app/Services/Beanstalkd.php
```diff
@@ -6,6 +6,7 @@ class Beanstalkd extends BaseService
 {
     protected $organization = 'schickling';
     protected $imageName = 'beanstalkd';
+    protected static $category = 'cache';
```
mattstauffer, line 9: Can we make all of these constants on a `Category` class so we can ensure they're consistent? So this line would instead be `protected static $category = Category::CACHE` or something similar

app/Services/BaseService.php
```diff
@@ -92,6 +93,11 @@ public function imageName(): string
         return $this->imageName;
     }
 
+    public static function category(): string
+    {
+        return Str::title(static::$category) ?? 'Other';
```
mattstauffer, line 98: Could we actually just return the lowercase name here, and then run `Str::title` on it in the consuming code?

app/Commands/EnableCommand.php
danielebarbaro: Thanks for suggestion.

app/Services/Beanstalkd.php
danielebarbaro: Yes also for me is better. In my private projects, I usually create a constant class to manage this strings.
So I created an abstract class Category placed in Services folder. Super clean.

danielebarbaro:
Hi @mattstauffer, thanks for your review. A made some review and a merge main in my branch to avoid conflicts.

app/Commands/EnableCommand.php
```diff
@@ -47,7 +59,20 @@ public function enableableServices(): array
         })->toArray();
     }
 
-    public function enable(string $service): void
+    public function enableableServicesByCategory(): array
```
mattstauffer, line 62: One last request! Can we sort the list after we're done so the category list is alphabetized?

danielebarbaro: Category sorted. 💪

mattstauffer:
I like it! Thanks @danielebarbaro! I'll let a few folks try it out before we tag a release.

PR #136, opened 2020-09-24
Add Linux, Windows Testing Again

### LINKED ISSUE #120: Update GitHub Actions text matrix to test on Linux and Windows again
Before going public the matrix was reduced to only test on macos to save build hours but as we're now public we could test on windows and linux again too as the build hours are unlimited for public repositories.

faxblaster:
Closes #120


mattstauffer:
Thanks @faxblaster!!

PR #142, opened 2020-09-29
Update build instructions in readme to be more explicit

delta1186:
This may seems like a trivial detail, but for someone who isn't familiar with compiling a build this might save them some time.  It wasn't abundantly clear to me that you should prefix the two steps in build instructions with `php `.  Thanks for your time and consideration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Make sure the `~/.composer/vendor/bin` directory is in your system's "PATH".
 
 ## Usage
 
-Run `takeout` and then a command name from anywhere in your terminal. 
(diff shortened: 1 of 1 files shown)


mattstauffer:
@delta1186 You shouldn't have to do that, because of this line at the top:

image

What system are you running on?

delta1186:
Windows 10

mattstauffer:
Ahhhh... freaking windows :D ok, we'll merge this in for the Windows folks! Thanks!

delta1186:
@mattstauffer thanks! I apologize on behalf of Windows. 😂🤣

PR #147, opened 2020-09-30
upgrade to guzzle 7

mihaileu:


josecanhelp:
Reading the upgrade guide, this seems like a safe upgrade. https://github.com/guzzle/guzzle/blob/7.0.0/UPGRADING.md#60-to-70

Issue #143, opened 2020-09-30
Add category to service list JSON


mattstauffer:
(See #88 and #135)

danielebarbaro:
I was thinking about json output days ago, feel free to assign also this to me 🤓.
I can easily handle this feature.
maybe we could talk about the right format

mattstauffer:
@danielebarbaro Sounds great! We actually already have JSON output, so this would be a quick one for you :)

mattstauffer:
https://github.com/tighten/takeout/pull/111

tonysm:
@mattstauffer I think we can close this one, right?

mattstauffer:
Yup!

mattstauffer:
Closed by #149.

PR #149, opened 2020-10-01
Add categories on json list

danielebarbaro:
Hi @mattstauffer , first implementation of #143 could be like this:

Screenshot 2020-10-01 at 17 30 16

The json is: 

```json
{
  "Cache": [
    [
      {
        "container_id": "3aac7ef4477f",
        "names": "TO--memcached--1.6.7",
        "status": "Up 7 hours",
        "ports": "0.0.0.0:11211->11211\\/tcp"
      }
    ]
  ],
  "Database": [
    [
      {
        "container_id": "4dc197902954",
        "names": "TO--mariadb--focal",
        "status": "Up 7 hours",
        "ports": "0.0.0.0:3342->3306\\/tcp"
      }
    ],
    [
      {
        "container_id": "0daed47e14a5",
        "names": "TO--postgresql--13.0",
        "status": "Up 3 days",
        "ports": "0.0.0.0:5647->5432\\/tcp"
      }
    ]
  ],
  "Mail": [
    [
      {
        "container_id": "202e824629b1",
        "names": "TO--mailhog--v1.0.1",
        "status": "Up 7 hours",
        "ports": "0.0.0.0:8025->8025\\/tcp, 0.0.0.0:2341->1025\\/tcp"
      }
    ]
  ]
}
```

I decided to group container by category as i've done in the principal menu #88 (PR #135) but we can discuss about it.

I think that isn't useful category column in list output.


danielebarbaro:
Hi @mattstauffer now PR is updated.
All tests pass.

This PR needs #157 .

danielebarbaro:
🥺

mattstauffer:
@danielebarbaro Whoops, sorry!

1. I *think* I'd prefer we use the lowercase version of the categories, but I'd be up to someone telling me the uppercase is better.
2. My natural inclination would be just to add it to each item in the list rather than grouping them this way. @mpociot as the first person who's talked about possibly consuming this data, what do you think?

danielebarbaro:
hi @mpociot , what you think about this implementation?

> I think I'd prefer we use the lowercase version of the categories, but I'd be up to someone telling me the uppercase is better.

oookkeeeei  🧐 🤓

> My natural inclination would be just to add it to each item in the list rather than grouping them this way. @mpociot as the first person who's talked about possibly consuming this data, what do you think?

I prefer grouping to "select" only one group a time 💪🏻

mattstauffer:
@danielebarbaro OK, I'm going to say definitely throw them into the main body of each entry, and give a flat response of entries.

@josecanhelp had a great idea as well: To add the category under a `meta` key, so it's separated from the rest.

Thanks!

danielebarbaro:
Cool, if you need any help i can support you .
I noticed your last PR merges and i saw **Base Alias** && **Full Alias** , well done @josecanhelp 🥇

mattstauffer:
@danielebarbaro If you're up to code the change, I'd love it! If you don't have time, please let me know.

danielebarbaro:
> @danielebarbaro If you're up to code the change, I'd love it! If you don't have time, please let me know.

i'm always on the cutting edge 🏄🏼‍♂️. i will rewrite the PR with the new changes

danielebarbaro:
Hi @mattstauffer , PR updated but i have some doubts to discuss.

Screenshot 2020-11-10 at 15 49 26


- `takeout list --json` have different output from `takeout list` it is what we want?

My opinion is to have same output (add category column on list) and with #173 filter base alias, full alias and category fields

- Full Alias and Base Alias are lowercase, category doesn't.

My opinion is to have same output :) convert category to lowercase and use ucfirst() on menu (`takeout enable`).

- behind the scenes i use container name to get the right category, i would like to use Base Alias

My opinion is: container name isn't clean to parse, base alias is "more confident" to use

danielebarbaro:
🙋🏻‍♂️

mattstauffer:
> takeout list --json have different output from takeout list is it what we want?
Yep!

> Full Alias and Base Alias are lowercase, category isn't.
I agree, let's strtolower the category

> behind the scenes i use container name to get the right category however, i would like to use base_alias
Makes sense!

app/Services/Category.php
```diff
@@ -10,4 +12,30 @@ abstract class Category
...
+    public static function all(): array
+    {
+        return collect((new Services)->all())
+            ->mapWithKeys(function ($fqcn, $shortName) {
+                return [$shortName => $fqcn::category()];
+            })
+            ->toArray();
+    }
+
+    public static function get(string $serviceName): string
```
mattstauffer, line 25: This would be `fromServiceName`, right?

app/Services/Category.php
```diff
@@ -10,4 +12,30 @@ abstract class Category
     const SEARCH = 'Search';
     const STORAGE = 'Storage';
     const TOOLS = 'Tools';
+
+    public static function all(): array
```
mattstauffer, line 16: I think this would more be a method on the Service class named `allByCategory` or something like that?

mattstauffer (changes requested):

app/Services/Category.php
danielebarbaro: Yep.

app/Services/Category.php
danielebarbaro: `all()` now returns:

```php
$result = [
    'beanstalkd' => "Cache",
    'clickhouse' => "Database",
    'dynamodb' => "Cache",
    'elasticsearch' => "Search",
    'eventstoredb' => "Database",
    'expose' => "Tools",
    'influxdb' => "Database",
    'mailhog' => "Mail",
    'mariadb' => "Database",
    'meilisearch' => "Search",
    'memcached' => "Cache",
    'minio' => "Storage",
    'mongo' => "Database",
    'mssql' => "Database",
    'mysql' => "Database",
    'neo4j' => "Database",
    'postgis' => "Database",
    'postgresql' => "Database",
    'redis' => "Cache",
    'sftp' => "Storage",
    'sqs' => "Cache"
];
```

so the right name (for me) is `mapByService()`

mattstauffer: Sure, but I still think this is more of a list of services than a list of categories, right?

mattstauffer:
@danielebarbaro Thanks for all your hard work on this!

app/Services/Category.php
danielebarbaro: whoops sorry 🙈 I misunderstood!
Apologies, you're right, it is a method for Service class.

mattstauffer:
@danielebarbaro OK, I'm finally getting back into my open source tasks. Taking a look now!

app/Services/Category.php
```diff
@@ -2,6 +2,8 @@
 
 namespace App\Services;
 
+use App\Services;
+
 abstract class Category
```
mattstauffer, line 7: I know this has been sitting open for ages so I'm sorry to make one more request, but could you write tests for these two new methods?

Thanks so much!

mattstauffer (approved):

tonysm:
Thanks @danielebarbaro for the work on this. And sorry for the long delay!

PR #153, opened 2020-10-03
Add tests for disable command

olivernybroe:
I have added some tests for the disable command, together with selecting things with the menu.

If the testing of menu is acceptable, then I wouldn't mind fixing the test marked `incomplete` also.

### DIFF
--- a/tests/Feature/DisableCommandTest.php
+++ b/tests/Feature/DisableCommandTest.php
@@ -2,81 +2,159 @@
 
 namespace Tests\Feature;
 
-use App\Commands\DisableCommand;
 use App\Shell\Docker;
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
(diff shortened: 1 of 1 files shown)


olivernybroe:
So I realised that while I created my pr, tests for the disable command was added in.

However those tests did not do anything at all, as it tried to mock the command itself which does not work.
So all of the expectations written in the mock of the command were useless. You can test it quickly by for example making tight expectations, like expecting that a certain method was only called once.

mattstauffer:
@olivernybroe Thanks for the updated note! Would you mind updating this PR so it doesn't have a merge conflict and then I'd be really happy to review it? Thanks so much!

olivernybroe:
@mattstauffer of course 👍 

I have removed the merge conflicts

mattstauffer:
Thanks @olivernybroe!

PR #154, opened 2020-10-05
Add support for EventStoreDB

morrislaptop:
+ some small fixes in the test class


tests/Feature/BaseServiceTest.php
```diff
@@ -42,10 +42,10 @@ function it_enables_services()
             $mock->shouldReceive('imageIsDownloaded')->andReturn(true);
 
             // This is the actual assertion
-            $mock->shouldReceive('bootContainer')->with(['getmeili/meilisearch']);
+            $mock->shouldReceive('bootContainer')->once();
```
mattstauffer, line 45: Why this replacement?

morrislaptop: Without specifying a call count the default is zero or more - http://docs.mockery.io/en/latest/reference/expectations.html#declaring-call-count-expectations

Tests were currently passing no matter the value of `bootContainer` or `getmeili/meilisearch`

mattstauffer: HA! Thanks @morrislaptop! Could you *keep* the original `with`?

morrislaptop: Hi @mattstauffer the original with isn't correct, it's going to have to be something like

```php
$mock->shouldReceive('bootContainer')->with('-p "${:port}":7700 \
            -v "${:volume}":/data.ms \
            "${:organization}"/"${:image_name}":"${:tag}"', [
                "organization" => "getmeili",
                "image_name" => "meilisearch",
                "port" => 7700,
                "tag" => "v1.1.1",
                "volume" => null,
                "container_name" => "TO--meilisearch--v1.1.1",
            ])->once();
```

which doesn't look very nice, any suggestions?

mattstauffer: Eh, let's just keep it as you have it for now and I'll dig in later to see if we really need that. Thanks so much!

mattstauffer:
@morrislaptop would you mind modifying this PR to allow me to make edits? I made some changes locally I'd like to push up; I'd like to push the changes up and then ask you to take a look at the failing tests.

Thanks!

morrislaptop:
@mattstauffer it was already checked but i've unchecked and checked it, does that work?

mattstauffer:
@morrislaptop Hm, nope, not sure why it isn't. Let me just throw you a patch file quickly.

mattstauffer:
Uh... I merged main in and pushed, which should've done almost nothing, and instead it's.. doing this? I'm not sure what's going on. :/

morrislaptop:
lol! Give me 5, I'll merge master in, there was only conflict on README IIRC

mattstauffer:
Thanks @morrislaptop! OK, so now where we are right now, we have two tests failures; would you mind taking a look?

image

morrislaptop:
@mattstauffer Tests and style fixed 👍

mattstauffer:
Thanks @morrislaptop!

Issue #162, opened 2020-10-13
Discussion: composer, composer.lock and php versions


danielebarbaro:
hi @mattstauffer and @josecanhelp your nightmare is coming, also today i spent some hours on takeout. 
The situation of composer and php version isn't clear at all for me 🤪.

With this composer.lock [94ec1a66498cd4f0e207afd3b40d8c2f6ab8e171] i can't install the project with php 7.2. The message is:

<img width="876" alt=" " src="https://user-images.githubusercontent.com/4376886/95825115-a7382d00-0d30-11eb-9b1b-e0e1a681601e.png">

If i drop composer.lock all things become nicer 😬

<img width="452" alt="Screenshot 2020-10-13 at 08 49 23" src="https://user-images.githubusercontent.com/4376886/95825440-23cb0b80-0d31-11eb-8683-3eabee8b5ca6.png">

We have this 8daa515251f58d11eddc002e0846c9e6f46aa8a6 commit but we have also this 94ec1a66498cd4f0e207afd3b40d8c2f6ab8e171 commit and i'm so confused that I created a new PR [#163] to remove composer.lock from repo.

Days ago, just for fun, i wrote this https://medium.com/@daniele.barbaro/the-composer-way-17ffa3205cfa to fight against this war 🤣

PR #163, opened 2020-10-13
Add composer.lock on gitignore and remove it from git cache

danielebarbaro:
Refs #162 💪


mattstauffer:
Hm. I get the value for this for the sake of individual contributors who aren't using PHP 7.4, but we're then also potentially introducing new problems with our builds where various contributors are building using different dependencies instead.

I think I may prefer requiring contributors to either A) use PHP 7.4 or B) only Git ignore `composer.lock` on their local machines so they update it to use their needed dependencies.

danielebarbaro:
As you know, different versions of composer may produce different behaviour. 
If we work on a project composer.lock makes the rule in a package i'm not sure at all.

Anyway, if this is your choice, maybe we can write the rules on:
- [Process for release](https://github.com/tighten/takeout#process-for-release) 
- [Contributing](https://github.com/tighten/takeout/blob/main/CONTRIBUTING.md) 

and 
 - reject all PR (except those of tighten team) with composer.lock changed
 - reject this PR 😖
 - close #162

Issue #165, opened 2020-10-16
Add SFTP server support


chrisdicarlo:
Hi,

I've built a custom service to allow spinning up SFTP servers using the atmoz/sftp image in my fork, just wondering if it would be of any interest to anyone for inclusion as a service in Takeout?

It currently supports:

- defining the port
- defining the SFTP user and password
- modifying the internal directory name on the server
- mapping a host directory into the container directly

I've got it working fine on my local takeout build.  More than happy to submit a PR if it's useful to anyone!

Cheers,
Chris

mattstauffer:
@chrisdicarlo Thanks for the issue! So this allows external folks to SFTP into your local machine, given your public IP? If so, this is definitely interesting!

chrisdicarlo:
@mattstauffer Actually, it was more for me to be able to spin up SFTP servers locally that can be used for testing, etc.  But your idea is interesting - just not sure how it could work consistently while still being local, i.e. without mucking about with any external firewall policies or port forwards.

I wonder if you could leverage Expose to tunnel it through somehow?  That would require a self-hoseted Expose server (another possible feature? :grin: ) but you could have the policies/forwards preset to point to it - but they would only be effective when you have actually spun up the Expose server?

chrisdicarlo:
Actually, I think even with Expose you would need to add a reverse proxy at the front end or a firewall policy / port forward anyway.  Either way will require some level of changes at the edge of the network.

If you have to do that anyway, you could simply have it point at the SFTP container directly.

PR #168, opened 2020-10-22
Add SFTP server support

chrisdicarlo:
Adds SFTP server support.  Does not allow direct outside communication on it's own but this could be accomplished via firewall policies/port forwarding.

Allows mapping of a local directory, a volume, or neither, via a prompt.  Related to #165

### DIFF
--- a/app/Services/Sftp.php
+++ b/app/Services/Sftp.php
@@ -57,7 +57,8 @@ class Sftp extends BaseService
         }, $this->defaultPrompts);
     }
 
-    protected function prompts(): void {
+    protected function prompts(): void
+    {
         parent::prompts();
 
         if($this->promptResponses['mapped_directory'] !== "") {
--- a/app/Services/Sftp.php
+++ b/app/Services/Sftp.php
@@ -2,11 +2,11 @@
 
 namespace App\Services;
 
-use App\Shell\Shell;
-use App\Shell\Docker;
-use App\Services\Category;
-use App\Shell\Environment;
 use App\Services\BaseService;
+use App\Services\Category;
+use App\Shell\Docker;
+use App\Shell\Environment;
+use App\Shell\Shell;
 
 class Sftp extends BaseService
 {
@@ -36,7 +36,7 @@ class Sftp extends BaseService
             'shortname' => 'mapped_directory',
             'prompt' => 'Which local directory should be mapped inside? (nothing if null)',
             'default' => '',
-        ]
+        ],
     ];
 
     protected $dockerRunTemplate = '-p "${:port}":22 \
--- a/app/Services/Sftp.php
+++ b/app/Services/Sftp.php
@@ -45,7 +45,8 @@ class Sftp extends BaseService
 
     protected static $displayName = 'SFTP';
 
-    public function __construct(Shell $shell, Environment $environment, Docker $docker) {
+    public function __construct(Shell $shell, Environment $environment, Docker $docker)
+    {
         parent::__construct($shell, $environment, $docker);
 
         $this->defaultPrompts = array_map(function ($prompt) {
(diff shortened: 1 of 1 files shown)


mattstauffer:
@chrisdicarlo From your issue:

> Actually, it was more for me to be able to spin up SFTP servers locally that can be used for testing, etc. 

I'm super curious about this. My biggest concern is us adding this and *not* supporting their public port forwarding out of the box; I think most folks would assume we're actually making this available publicly.

Just for me to understand it better, could you share an example of when you need to test it locally? Are you writing code that hits SFTP and this is just to test that code?

Thanks!

chrisdicarlo:
@mattstauffer I see your point about the public port forwarding; I don't really see any way to seamlessly get that to work without implementing some kind of 3rd-party server that tunnels the SFTP traffic over HTTPS to get around the firewall configuration issues.

That's exactly the use case I have - I have a some orchestration with various storage backends and this let's me spin up multiple SFTP servers locally to quickly test.  For my use case, external access isn't really a necessity.

chrisdicarlo:
One thing that kind of makes me feel it's not terrible to need some setup (if needed) is that takeout is primarily targeted at devs wanting to spin up containers quickly and easily - not really for production use.  Just my 2 cents though! :grin:

mattstauffer:
@chrisdicarlo OK, thanks--so I think it might seem like I'm really slow here, and I bet I am, but could you give me the practical use case for this again?

Basically, my question is:

In what context would it be useful for the average programmer to be able to run a locally-hosted SFTP server that doesn't connect to the outside world?

_Or_, is there no such case, and this is merely simplifying _one_ part of the process of them setting locally-hosted SFTP servers that *do* connect to the outside world?

Thanks!

chrisdicarlo:
The only scenario would be a developer who needs their application to interact with an SFTP server in some way and wants to be able to test and debug that locally without having to have a fully fledged external SFTP service running.￼

The external access is really secondary and not how I intended to use it.￼

mattstauffer:
@chrisdicarlo Oh, duh. I don't know why I didn't even think about it in terms of _making sure your application can correctly interact with SFTP in general._

I'm an idiot. Thanks!

app/Services/Sftp.php
```diff
@@ -0,0 +1,81 @@
...
+            }
+
+            return $prompt;
+        }, $this->defaultPrompts);
+    }
+
+    protected function prompts(): void {
+        parent::prompts();
+
+        if($this->promptResponses['mapped_directory'] !== "") {
```
mattstauffer, line 63: There are a few lines in here that I know don't work with the Tlint specs. Would you mind running TLint on this PR? Thanks!

https://github.com/tighten/tlint

chrisdicarlo: I just ran the linter and it just spits out **LGTM!**, which looks like there are no lints or it was successful?

chrisdicarlo: I fixed the issue withe position of the opening curly bracket.

mattstauffer: Hm. LGTM = Looks good to me, but this is definitely not tlint style. Let me try it.

mattstauffer: @chrisdicarlo Ahh, I mixed up. We *do* want to run Tlint, but this particular set of issues comes from the PSR's, so that'd come from running PHPCS. You can learn more in our contributing.md: https://github.com/tighten/takeout/blob/main/CONTRIBUTING.md

chrisdicarlo: My bad, I forgot to run it through phpcbf before I submitted the PR.  I just pushed a commit to fix things up.

app/Services/Sftp.php
```diff
@@ -0,0 +1,85 @@
...
+        "${:organization}"/"${:image_name}":"${:tag}" \
+        "${:user_config}"';
+
+    protected static $displayName = 'SFTP';
+
+    public function __construct(Shell $shell, Environment $environment, Docker $docker)
+    {
+        parent::__construct($shell, $environment, $docker);
+
+        $this->defaultPrompts = array_map(function ($prompt) {
```
mattstauffer, line 52: This is a very clean way to modify it. Nice work!

mattstauffer (approved):

PR #170, opened 2020-11-06
Add networking support

mattstauffer:
Networking Todo:

- [x] Put all containers into a `takeout` network
- [x] Test (somehow?) that this network means other Docker containers can be accessed by their full name (e.g. `TO--mailhog--v1.0.1`) in the target Docker-based Laravel apps
- [x] Figure out how to make it so my MySQL Docker container is accessible as `mysql`--alias
- [x] How do we make it so people with only one container get `mysql` easily, but people with multiple MySQL containers also have something usable
- [x] Tag all new containers with serviceVersion (e.g. `mysql80`)
- [x] Tag the first container with the short version (e.g. `mysql`)
- [x] Clean up the version number?
- [x] Extract Docker Networking methods to a separate class
- [x] Test the things with `@test` on them
   
Brainstorm on the naming situation:
- By default, each service aliases itself to just its service name (e.g. mysql)

**Every container gets tagged with `mysql80`**:

@josecanhelp gave his blessing to Option 1 🙌 

- Option 1:
    - All MySQL instances (e.g.) are tagged with their name and version (e.g. `mysql80`) *and* the first instance of that type that I spin up also gets the short version (`mysql`), with the intention of *later* making it so we can allow them to repoint the shortcut

- Option 2:
    -  If I spin up a second MySQL instance, it spins down my first, and then names them all as mysql57, mysql80, etc. *and* points the short name (`mysql`) to the _latest_ of those I have open

**Every container doesn't necessarily get tagged with the version**:
- Option 3:
    -  If I spin up a second MySQL instance, it spins down my first, and then names them all as mysql57, mysql80, etc.

- Option 4: (not prefered)
    - We require a nickname for every new container spun up, and that's used for its alias

- Option 5: (not prefered)
    - Number them. `mysql1`, `mysql2` ,
    etc


mattstauffer:
Not sure why WIndows is erroring but it has nothing to do with this code.

mattstauffer:
Crap. One fix needed:

image

This thing is way too long.

PR #171, opened 2020-11-08
Add SQS support

pmayet:
Add SQS support

- Default Node Host Port : 9324
- Management Http Host Port : 9325


mattstauffer:
Uh.. wow. I had no idea this existed!!

PR #176, opened 2020-11-09
Add support for Neo4j

zoispag:
Add support for [Neo4j](https://neo4j.com/)

- Default port: 7474
- Bolt access port: 7687

### DIFF
--- a/tests/Feature/ServicesTest.php
+++ b/tests/Feature/ServicesTest.php
@@ -28,7 +28,7 @@ class ServicesTest extends TestCase
         $filtered = collect($services->all())->filter(function ($fqcn, $shortname) {
             return Str::contains($fqcn, 'App\Services');
         })->filter(function ($fqcn, $shortname) {
-            return !! preg_match('/^[a-z]+$/', $shortname);
+            return !! preg_match('/^[a-z0-9]+$/', $shortname);
         });
 
         $this->assertEquals($all->count(), $filtered->count());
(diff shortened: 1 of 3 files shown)


tests/Feature/ServicesTest.php
```diff
@@ -28,7 +28,7 @@ function all_array_is_keyed_by_shortnames_and_values_are_fqcns()
         $filtered = collect($services->all())->filter(function ($fqcn, $shortname) {
             return Str::contains($fqcn, 'App\Services');
         })->filter(function ($fqcn, $shortname) {
-            return !! preg_match('/^[a-z]+$/', $shortname);
+            return !! preg_match('/^[a-z][a-z0-9]+$/', $shortname);
```
mattstauffer, line 31: I'm not a big regex guy, but is the first `[a-z]` block necessary now that your second block has `a-z` and `0-9`?

zoispag: My original change was `'/^[a-z0-9]+$/'`, but then it would allow an FQDN to start with a digit. I have no strong preference, but I was not sure if starting with a digit should be "allowed". (whatever the definition of allowed is). Let me know if you want me to change it :)

mattstauffer: Oo, good call! No, this is great. PHP classes can't start with a digit, so while this is unnecessary, it is more accurate.

PR #181, opened 2020-11-10
Hotfix/reorder services alphabetically

danielebarbaro:
After add sftp on PR #180 i realised that services wasn't ordered


mattstauffer:
Thanks!!

PR #183, opened 2020-11-10
Add container log command.

danielebarbaro:
For lazy people a new log command.

### DIFF
--- a/app/Commands/LogCommand.php
+++ b/app/Commands/LogCommand.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace App\Commands;
+
+use App\InitializesCommands;
+use App\Shell\Docker;
+use Illuminate\Support\Str;
+use LaravelZero\Framework\Commands\Command;
+
+class LogCommand extends Command
+{
+    use InitializesCommands;
+
+    protected $signature = 'logs {containerId?}';
+    protected $description = 'Display container logs.';
+
+    public function handle(): void
+    {
+        $this->initializeCommand();
+
+        $container = $this->argument('containerId');
+
+        if (! $container) {
+            $this->error("Please insert a valid container ID.\n");
+
+            return;
+        }
+
+        $this->logs($container);
+    }
+
+    public function logs(string $container): void
+    {
+        if (Str::contains($container, ' -')) {
+            $container = Str::before($container, ' -');
+        }
+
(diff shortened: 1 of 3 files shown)


danielebarbaro:
🙋🏻‍♂️

app/Commands/LogCommand.php
```diff
@@ -0,0 +1,40 @@
...
+            return;
+        }
+
+        $this->logs($container);
+    }
+
+    public function logs(string $container): void
+    {
+        if (Str::contains($container, ' -')) {
+            $container = Str::before($container, ' -');
```
mattstauffer, line 35: What's this covering?

app/Shell/Docker.php
```diff
@@ -46,6 +46,21 @@ public function stopContainer(string $containerId): void
         }
     }
 
+    public function logContainer(string $containerId): void
+    {
+        if (! $this->stoppableTakeoutContainers()->contains(function ($container) use ($containerId) {
+            return $container['container_id'] === $containerId;
+        })) {
+            throw new DockerContainerMissingException($containerId);
```
mattstauffer, line 54: Does this give a clear error? Or should we catch it in the command and present a cleaner error?

mattstauffer (changes requested):
Thanks for the PR! Just a few questions

app/Commands/LogCommand.php
danielebarbaro: Good question.
When I created the command I copied startCommand class and I forgot to delete that rows.
Is still necessary that rows in start and stop commands?

app/Shell/Docker.php
danielebarbaro: For me is super clear and it isn't necessary a specific exception 🙂. 
The message is: "Docker container 123456 not found." and it is correct because you can log only an active container 
If you feel more comfortable I can provide a specific exception.

mattstauffer: Nope, that's great! I just hadn't tested it :)

app/Commands/LogCommand.php
mattstauffer: I honestly don't remember what it was for :)

danielebarbaro: me too🤣 but now it is ready to merge.

PR #185, opened 2020-11-17
Use more specific term for interface and rename types file

josecanhelp:
Typescript convention uses `types.ts` to store interfaces.
I thought using `DockerContainer` instead of `Container` would make our code more easily readable.


mattstauffer:
Love it--thanks! :)

PR #192, opened 2020-12-03
Add --all in start command

chrysanthos:
This will allow us to start all containers at once without pressing enter 5 times or even add it to our .bash_profile so that they start automatically when opening the command line.

### DIFF
--- a/app/Commands/StartCommand.php
+++ b/app/Commands/StartCommand.php
@@ -19,10 +19,12 @@ class StartCommand extends Command
     {
         $this->initializeCommand();
 
-        $container = $this->argument('containerId');
+        $containers = $this->argument('containerId');
 
-        if ($container) {
-            $this->start($container);
+        if ($containers) {
+            foreach ($containers as $container) {
+                $this->start($container);
+            }
 
             return;
         }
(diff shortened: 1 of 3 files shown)


mattstauffer:
I love this.

app/Commands/StartCommand.php
```diff
@@ -12,7 +12,7 @@ class StartCommand extends Command
 {
     use InitializesCommands;
 
-    protected $signature = 'start {containerId?}';
+    protected $signature = 'start {containerId?*} {--all}';
```
mattstauffer, line 15: What's the `*`? If it captures *anything* passed (so someone could pass more than one), wouldn't we need to update the consuming code to handle that? Thanks!

chrysanthos: Initially added the asterisk because `takeout start --all` would still consider `--all` to be a `containerId` and fail. Not sure what I did wrong there. 

You are right though, I will update my PR to handle multiple containers.

markokeeffe:
Would love to use this feature!

mattstauffer:
Sorry, I didn't see that @chrysanthos had updated the code in response to my note. I'm reading it again now, thanks!

Issue #195, opened 2020-12-11
Cannot stop an already stopped container


delta1186:
When running the disable command the stopContainer method is called inside of removeContainer.  If the container is not currently running it will throw an exception causing the removeContainer method to end before actually removing the container.  

I have written a PR to fix this issue, which just wraps the stopContainer method call inside a conditional.  This is the same conditional used inside the stopContainer method which throws the DockerContainerMissingException.   I guess the other way you could fix this is just wrap the stop logic inside the conditional inside the stopContainer method and just remove the section where the exception is thrown.  

If that method would be preferred, then I can update my PR accordingly.  Thanks!

delta1186:
Hey @mattstauffer this one should be pretty straight forward.  Just added a test for the issue I fixed in PR #196. 

Just let me know if you have any questions or concerns. This would close this issue.

PR #196, opened 2020-12-11
Fix bug cannot stop an already stopped container

delta1186:
Fixes issue #195 

See commits in the issue.  If I need to make any modifications, then just let me know.  Thanks!


mattstauffer:
Thanks @delta1186!!

PR #197, opened 2020-12-14
Add test to validate fix for issue 195

delta1186:
Ok in an attempt to be thorough and try and work on getting better at testing, I created a test that validates the fix that was merged in PR #196.  This should close issue #195.  

If you comment out the conditional that was merged already, then this new test will fail.  So, I think it is validating things correctly.

Let me know if you need me to make any changes.  Thanks!


mattstauffer:
Thanks @delta1186 !

Issue #200, opened 2020-12-22
Dependency Injection


delta1186:
I am wondering if the start, stop, enable, and disable commands should be standardize with some injected dependencies.  This will avoid some work arounds with testing I had to implement in order to handle a class is not tied to the command, but rather directly loaded from the container.  Hope I am saying that right. :)  Like the following example from line [63](https://github.com/tighten/takeout/blob/0ac5975aac8843330428163267acb60b5d609552/app/Commands/StartCommand.php#L63) in StartCommand.

`app(Docker::class)->startContainer($container)`

I would like to implement constructor injection for Docker, Environment, and also Services for some commands.  It is already being done to some scale in the commands.  Some something like this.

```
   protected $description = 'Start a stopped container.';
    protected $docker;
    protected $environment;

    public function handle(Docker $docker, Environment $environment): void
    {
        $this->environment = $environment;
        $this->docker = $docker;
        $this->initializeCommand();
```

This would prevent me from having to add the following to some of the tests to get them to run in all of the required php versions.

```
/** @test
     * @runInSeparateProcess
     * @preserveGlobalState disabled
     */
```

@mattstauffer or @josecanhelp let me know if you are okay with this and I can make a PR for it.  

This won't stop me from finishing my tests for Windows menus, so I am fine either way.  If you do want to do this later, then I can just refactor my tests if that PR is already merged as part of this PR.  Just wanted to make the suggestion now.

mattstauffer:
@delta1186 Sounds good to me! Thanks!!

delta1186:
Hey @mattstauffer I am working on a PR for this and before submitting I was trying to figure out why your Actions were failing for php7.2.  I appears that PHPUnit 9 that you moved to no longer supports Php 7.2.  Do you want to remove 7.2 support or rollback PHPUnit?

mattstauffer:
@delta1186 Thanks--I'd like to eventually bring that support back in via a PHPUnit bridge command, but you can comment out 7.2 support for now. Thanks!

PR #201, opened 2020-12-28
Inject dependencies into constructors

delta1186:
This PR resolves issue #200 

I have updated the commands to include some injected dependencies.  I also removed php 7.2.5 support due to takeout being updated to use phpunit 9.5 and laravel zero 8.0 which require ph 7.3 and higher.  Had to add the following to composer.json in order to get it to update in Windows. :)

```
"platform-check": false,
        "platform": {
            "ext-pcntl": "7.2",
            "ext-posix": "7.2"
        }
```

Let me know if you need me to make any changes.


composer.json
```diff
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": "^7.2.5 || ^8.0",
+        "php": "^7.3 || ^8.0",
```
mattstauffer, line 19: I _believe_ Takeout still works fine on PHP 7.2; it's only the tests that don't. As such, I'd rather we keep it at `7.2` since I still intend to fix the tests.

Thanks!

app/Commands/EnableCommand.php
```diff
@@ -13,9 +15,15 @@ class EnableCommand extends Command
 
     protected $signature = 'enable {serviceNames?*} {--default}';
     protected $description = 'Enable services.';
+    protected $docker;
+    protected $environment;
+    protected $services;
 
-    public function handle(): void
+    public function handle(Docker $docker, Environment $environment, Services $services): void
```
mattstauffer, line 22: No need to inject Docker and Environment, since we're not using them.

app/Commands/ListCommand.php
```diff
@@ -12,20 +12,24 @@ class ListCommand extends Command
 
     protected $signature = 'list {--json}';
     protected $description = 'List all services enabled by Takeout.';
+    protected $docker;
 
-    public function handle(): void
+    public function handle(Docker $docker): void
     {
+        $this->docker = $docker;
```
mattstauffer, line 19: Since it's all in the `handle` method, let's just keep it simple and make the call directly on the `$docker` object and skip assigning it as a class property.

app/Commands/LogCommand.php
```diff
@@ -13,9 +13,11 @@ class LogCommand extends Command
 
     protected $signature = 'logs {containerId?}';
     protected $description = 'Display container logs.';
+    protected $docker;
 
-    public function handle(): void
+    public function handle(Docker $docker): void
     {
+        $this->docker = $docker;
```
mattstauffer, line 20: Skip assigning it as a class property, please

mattstauffer (changes requested):
Looks great! Just a few small changes.

delta1186:
Hey @mattstauffer got your requests. I will implement those changes first thing in the morning when I get to work. Sorry I got a little ahead of myself on some of those dependency injections. Using them in my windows menu PR. 😅

Thanks for the feedback! Feel like I'm learning a lot from contributing to the open source community. 🤙🏻￼

app/Commands/LogCommand.php
delta1186: Hey @mattstauffer so I left this one because it is used in the logs function in that class.

mattstauffer:
Thanks so much @delta1186! :)

PR #202, opened 2021-01-05
Add RabbitMQ service

bandgeekndb:
Adds RabbitMQ as a service. This is an available feature on Laravel Homestead and one I use for my development, so I'd like to see it available in Takeout if possible!

Notes:

- RabbitMQ relies on the hostname to organize data, so it's important we specify the hostname (-h option, comes from [RabbitMQ docker docs](https://hub.docker.com/_/rabbitmq))
- The default `latest` tag does not enable the management plugin, so opening/asking about the management port is unnecessary. I didn't know any way to conditionally ask for a management port based on the tag selected, so I overrode the default prompts so that `management` is the suggested default tag. Not sure if that's bad form 😂

Thanks for considering this! Love Takeout and how easy it makes using Docker in my dev workflow!

### DIFF
--- a/app/Services/RabbitMQ.php
+++ b/app/Services/RabbitMQ.php
@@ -8,6 +8,18 @@ class RabbitMQ extends BaseService
 
     protected $imageName = 'rabbitmq';
     protected $defaultPort = 5672;
+    protected $defaultPrompts = [
+        [
+            'shortname' => 'port',
+            'prompt' => 'Which host port would you like %s to use?',
+            // Default is set in the constructor
+        ],
+        [
+            'shortname' => 'tag',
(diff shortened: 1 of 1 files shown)


mattstauffer:
Thanks! Is the `management` branch relatively up-to-date, just with an extra feature turned on? If so, I think that's fine!

bandgeekndb:
Hi Matt! I just double-checked the docs, the `management` tag is a mirror of the `latest` tag, just with the management plugin enabled, so I think we can safely use the tag and stay up-to-date. Thanks!

mattstauffer:
Brilliant, thanks!

PR #206, opened 2021-01-13
Add matrix entry for Apple Silicon

### LINKED ISSUE #205: Add PHP8-AppleSilicon case to test Matrix
With the release of Apple Silicon we want to make sure that the application behaves properly, `shivammathur/setup-php` currently supports only PHP 8 on Apple Silicon. I propose adding this test case to the suite.

abenettt:
Closes #205 

This PR adds a matrix entry to also test `takeout` on Apple Silicon, this way all platforms that `takout` is probably going to be run on are covered.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -23,6 +23,10 @@ jobs:
                   - os: macOS
                     os-version: macos-latest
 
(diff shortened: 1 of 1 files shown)


mattstauffer:
Thanks!!

PR #210, opened 2021-01-21
Fixes PSR-4 issues

mabadir:
Fixes the below error during `composer install`

```
Class Tests\Feature\MicrosoftDockerTagsFakestream located in ./tests/support/MicrosoftDockerTagsFakestream.php does not comply with psr-4 autoloading standard. Skipping.
```


mattstauffer:
Thanks @mabadir!

mabadir:
First, but I hope not the last PR...

PR #212, opened 2021-01-31
add MailDev service

godbout:
this PR adds the MailDev service as an alternative to MailHog.

why is MailDev MegaCool™?

1. when you get a new email you're getting updated directly in the tab title

Screen Recording 2021-02-01 at 00 25 38

2. MailDev has some PrettyCool™ features like responsive view, see email as text, etc

Screen Recording 2021-02-01 at 00 29 14

might be worth adding some competition to MailHog. who knows.

### DIFF
--- a/app/Services/MailDev.php
+++ b/app/Services/MailDev.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace App\Services;
+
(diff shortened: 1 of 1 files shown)


mattstauffer:
Thanks @godbout!

Issue #211, opened 2021-01-28
change in `takeout start`? can't start multiple items from the menu


godbout:
i'm pretty sure in the past i would run `takeout start` and be able to start several containers. i would choose one, it would disappear from the list, choose another one, etc..., until i choose exit. now the menu quits right after my first choice. i've checked the docs to see if something has changed but even there's some new features, there's nothing that should have affected this. am i getting old? :D

godbout:
same for `takeout stop` btw.

mattstauffer:
Ahh... a PR was sent in to update the menu to work on Windows, and they may have unintentionally broken this flow for Mac users. We'll take a look--thanks!

delta1186:
@mattstauffer I will take a look at this since that was my PR.  If my memory serves me, only the disable menu had that functionality which is what I replicated on the Windows side.  Let me checkout an older commit and review the behavior.

godbout:
@delta1186 i don't use disable. enable once, then each time i work on a project, i `takeout start` and `takeout stop`. i remember starting and stopping a couple of containers in one go. navigating with `jk` to get to the right containers, start/stop, then exit.

mattstauffer:
@delta1186 Thanks! Yah, the `start` menu and I *think* the `stop` menu used to let you keep starting/stopping until you were done.

delta1186:
Sorry I am just looking at this.  Not sure how I broke it, but you are correct that it is not working.  However, it probably was working, because I replicated the behavior in the Window's menu.  So currently it works on Windows for start and stop, but Linux (and I assume Mac) it does not.  Hopefully it should be an easy fix.

mattstauffer:
Just to clarify, @delta1186, does that mean I should pick up this fix, or is it something you'll be looking at? No worries if you can't, just trying to clarify :)

delta1186:
@mattstauffer no no I was just explaining what I saw￼. I should be able to fix it easy enough this morning. 🤙🏻￼

delta1186:
Okay @mattstauffer should be good to go. :)

mattstauffer:
Fixed in #214; thanks so much @delta1186!!

godbout:
thanks you guys!

PR #214, opened 2021-02-09
Fix start and stop command bug where default menu closes early

delta1186:
This should resolve issue #211.  Apparently my original implementation to close the start and stop command menu if only the exit item remained did not work for the default menu layout.  This has been reworked and now works as intended.


mattstauffer:
Thanks so much @delta1186!

PR #220, opened 2021-03-24
Revise error messages to include Linux debugging help

benholmen:
I just set up Takeout on Ubuntu 20.04 and I'm really pleased - thank you! I ran into a couple of errors and have some suggestions to provide more helpful error messages to Linux users.


app/Exceptions/DockerNotAvailableException.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace App\Exceptions;
+
+use App\Shell\Environment;
+use App\Shell\Shell;
+use Exception;
+
+class DockerNotAvailableException extends Exception
```
benholmen, line 9: I found that even though Docker was installed, it wasn't available to a non-root user without some extra steps. I renamed the exception to reflect that distinction.

app/Exceptions/DockerNotAvailableException.php
```diff
@@ -0,0 +1,45 @@
...
+        $console->line($shell->formatErrorMessage('Docker is not available.'));
+        $console->line('');
+
+        if ($environment->isLinuxOs()) {
+            $console->line('Verify that the Docker service is running:');
+            $console->line('  sudo systemctl status docker');
+            $console->line('Start the Docker service:');
+            $console->line('  sudo systemctl start docker');
+
+            if (! $environment->userIsInDockerGroup()) {
```
benholmen, line 27: I'm not crazy about this logic in the exception. However, I decided not to throw a unique exception when a user is not in the docker group because it's possible that some environments don't require it. I wanted this message to be a suggestion for why the `docker info` command is failing, not a requirement to use Takeout.

app/Exceptions/DockerNotAvailableException.php
```diff
@@ -0,0 +1,45 @@
...
+        $console = app('console');
+        $shell = app(Shell::class);
+        $environment = app(Environment::class);
+
+        $console->line('');
+        $console->line($shell->formatErrorMessage('Docker is not available.'));
+        $console->line('');
+
+        if ($environment->isLinuxOs()) {
+            $console->line('Verify that the Docker service is running:');
```
mattstauffer, line 22: Could we move Linux, Windows, and Mac into their own protected methods? That'll start to make this a bit more manageable.

app/Exceptions/DockerNotAvailableException.php
mattstauffer: Totally get that.

mattstauffer (changes requested):
Great work on this, thanks!

app/Exceptions/DockerNotAvailableException.php
```diff
@@ -0,0 +1,63 @@
...
+        $shell = app(Shell::class);
+
+        $console->line('');
+        $console->line($shell->formatErrorMessage('Docker is not available.'));
+        $console->line('');
+
+        if (in_array(PHP_OS_FAMILY, ['Darwin', 'Linux', 'Windows'])) {
+            $osSpecificHelp = 'helpFor' . ucfirst(PHP_OS_FAMILY);
+            $this->$osSpecificHelp();
+        }
```
benholmen, line 23: @mattstauffer I'm not sure this is much cleaner but it's at least separated by environment. I'm open to any suggestions!

app/Exceptions/DockerNotAvailableException.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    protected function helpForDarwin()
+    {
+        $console = app('console');
```
benholmen, line 28: @mattstauffer is there a way to avoid resolving this in every method?

app/Exceptions/DockerNotAvailableException.php
mattstauffer: I like it!

app/Exceptions/DockerNotAvailableException.php
```diff
@@ -0,0 +1,63 @@
...
+        $shell = app(Shell::class);
+
+        $console->line('');
+        $console->line($shell->formatErrorMessage('Docker is not available.'));
+        $console->line('');
+
+        if (in_array(PHP_OS_FAMILY, ['Darwin', 'Linux', 'Windows'])) {
+            $osSpecificHelp = 'helpFor' . ucfirst(PHP_OS_FAMILY);
+            $this->$osSpecificHelp();
```
mattstauffer, line 22:
```suggestion
            $this->$osSpecificHelp($console);
```

app/Exceptions/DockerNotAvailableException.php
mattstauffer: Yep! As with my suggestions, pass `$console` into the method call, and then accept it in each function definition

app/Exceptions/DockerNotAvailableException.php
```diff
@@ -0,0 +1,63 @@
...
+            $console->line('Add your user to the docker group by running:');
+            $console->line('  sudo usermod -aG docker ${USER}');
+            $console->line('and restart your session.');
+        }
+    }
+
+    protected function helpForWindows()
+    {
+        $console = app('console');
+
```
mattstauffer, line 60:
```suggestion
    protected function helpForWindows($console)
    {
```

app/Exceptions/DockerNotAvailableException.php
```diff
@@ -0,0 +1,63 @@
...
+        $console->line('Open Docker for Mac or run:');
+        $console->line('  open --background -a Docker');
+        $console->line('to start the Docker service.');
+    }
+
+    protected function helpForLinux()
+    {
+        $console = app('console');
```
mattstauffer, line 37:
```suggestion
    protected function helpForLinux($console)
    {
```

app/Exceptions/DockerNotAvailableException.php
mattstauffer, line 28:
```suggestion
    protected function helpForDarwin($console)
    {
```

mattstauffer (changes requested):
Just quick formatting changes, then we're good. Thanks!

mattstauffer:
Thanks @benholmen!

PR #222, opened 2021-04-09
Adds port availability check for the MySQL service

tonysm:
### Added

- Adds the port availability check for the MySQL service. I think this is something we could extract (to a validators folder?) so we could reuse in other services
- Adds a mention to install the `netstat` CLI tool if it's not already installed by default in the distro (mine wasn't)

### DIFF
--- a/src/shell/environmentshell.ts
+++ b/src/shell/environmentshell.ts
@@ -7,11 +7,11 @@ export default class EnvironmentShell {
   }
 
   static netstatCmd(port: number): any {
-    // @TODO: is there any danger in concatenating the command like this? Do we need to escape it somehow?
     const portText = EnvironmentShell.isLinuxOs()
       ? `:${port} `
       : `.${port} `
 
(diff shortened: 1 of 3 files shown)


src/shell/environmentshell.ts
```diff
@@ -1,7 +1,23 @@
...
+    const result = EnvironmentShell.netstatCmd(port)
+    return result.status !== 0
+  }
+
+  static netstatCmd(port: number): any {
+    const portText = EnvironmentShell.isLinuxOs()
+      ? `:${port} `
+      : `.${port} `
+
+    // @TODO: is there any danger in concatenating the command like this? Do we need to escape it somehow?
```
tonysm, line 14: @josecanhelp is this a problem? In PHP we usually use `escapeshellarg` when adding arguments to command like this. Didn't find references to things like that in NodeJS, so I wasn't sure.

josecanhelp (approved):

PR #225, opened 2021-04-25
Added TimescaleDB service and fixed prompts

octoper:
Added [TimescaleDB](https://www.timescale.com/) as a service.

Fixed the `prompts()` function in `BaseService` to merge `$prompts` and `$defaultPrompts` and mapped all the prompts by their short-name so there will not be duplicates be `$prompts` and `$defaultPrompts`

### DIFF
--- a/app/Services/TimescaleDB.php
+++ b/app/Services/TimescaleDB.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace App\Services;
+
+class TimescaleDB extends BaseService
+{
+    protected static $category = Category::DATABASE;
+
+    protected static $displayName = 'TimescaleDB';
+
+    protected $organization = 'timescale';
+    protected $imageName = 'timescaledb';
+    protected $defaultPort = 5432;
+
+    protected $prompts = [
+        [
+            'shortname' => 'volume',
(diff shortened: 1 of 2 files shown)


app/Services/TimescaleDB.php
```diff
@@ -0,0 +1,36 @@
...
+        ],
+        [
+            'shortname' => 'root_password',
+            'prompt' => 'What will the password for the `postgres` user be?',
+            'default' => 'password',
+        ],
+        [
+            'shortname' => 'tag',
+            'prompt' => 'Which tag (version) of %s would you like to use?',
+            'default' => '2.2.0-pg13',
```
mattstauffer, line 29: I don't think we want to default to a fixed version, because then we'll have to constantly keep updating this.

Is there a reason the tag our default tag picker returns is incorrect?

In that case, we may need to instead provide a new default tag picker just for this service.

octoper: Sorry there was not a need for these changes, I just had an issue with the latest version and it did't pulled the latest image correctly I'm gonna close this PR and create a new one with just the TimescaleDB service.

PR #226, opened 2021-04-30
Fix docker tags API response changed

tonysm:
### Fixed

- Fix the `DockerTagsTest` because the API response changed from the DockerHub API

---

Not sure if this is the most "correct" fix, though.


mattstauffer:
Thanks @tonysm!

PR #227, opened 2021-05-01
Added TimescaleDB service

octoper:
Added [TimescaleDB](https://www.timescale.com/) as a service.

### DIFF
--- a/app/Services/TimescaleDB.php
+++ b/app/Services/TimescaleDB.php
@@ -25,7 +25,8 @@ class TimescaleDB extends BaseService
         ],
     ];
 
-    protected $dockerRunTemplate = '-p "${:port}":9000 \
+    protected $dockerRunTemplate = '-p "${:port}":5432 \
+        -e POSTGRES_PASSWORD="${:root_password}"\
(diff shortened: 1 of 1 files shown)


mattstauffer:
When trying to run this, I got this error in my Docker logs:

image

octoper:
Hey @mattstauffer i fixed that error but now I have came up to an error I have faced before https://github.com/tighten/takeout/pull/225 the default TimescaleDB tag is `latest-pg12` but in the prompts the default tag is `latest`

mattstauffer:
@octoper OK, we actually just merged in something (it's not tagged yet) that allows you to use custom images, and since we already have a Postgres installation, you could potentially just use that instead of adding Timescale.

Would you test this out for me? Just update your local install to the latest version of `main`, then do this:

image

(You can use whatever port you want)

It said it was successful for me, but you're the best to test out whether it's working the way you want.

riasvdv:
This is failing for me, it doesn't seem to parse the custom tag correctly.

image

riasvdv:
Seems to be an issue on my end, added a test in #247 which verifies it works.

PR #239, opened 2021-07-27
Add a static port for the MinIO Console.

SebastianBerc:
Hello, by default, MinIO Console is listening on a dynamic port that is randomly generated in every run, which makes it impossible to log in.

> WARNING: Console endpoint is listening on a dynamic port (42055), please use --console-address ":PORT" to choose a static port.

https://docs.min.io/docs/minio-docker-quickstart-guide.html

### DIFF
--- a/app/Services/Minio.php
+++ b/app/Services/Minio.php
@@ -16,6 +16,11 @@ class Minio extends BaseService
             'default' => 'minio_data',
         ],
         [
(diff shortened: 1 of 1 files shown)


therobfonz:
This would be great to get merged. Seems like newer version of Minio this happens.

mattstauffer:
Thanks for the PR!

PR #257, opened 2021-11-05
Upgrade Composer

tonysm:
### Changed

- Bumps composer dependencies (ran `composer update -W`)
- Bumps minimum PHP version to 7.3, we're not testing against 7.2 on GH Actions and the PHPUnit version we're currently using requires 7.3, so I think requiring 7.2 just opens up for more unforeseen problems on our end

### Fixed

- Fixed a missing type-hint in the `MinioDockerTagsFakestream` (which extends `GuzzleHttp\Psr7\Stream`)
- Fixes the PSR-2 autoloading warning, which was forcing us to manually require some Support classes in some test cases.


mattstauffer:
Thanks @tonysm!

PR #256, opened 2021-11-05
Use mysql/mysql-server image for ARM processors

tonysm:
### Changed

- Changes the MySQL service to use the `mysql/mysql-server` image by default instead of the `mysql` one, as the former seems to work well on ARM-based processors (M1)

---

This is based on a recent change made to [Laravel Sail](https://github.com/laravel/sail/pull/272)

PR #258, opened 2021-11-08
Revert "Use mysql/mysql-server image for ARM processors"

mattstauffer:
Reverts tighten/takeout#256


mattstauffer:
#256 wasn't ready and I accidentally merged it.

Issue #190, opened 2020-11-30
MySQL said: Host '172.18.0.1' is not allowed to connect to this MySQL server


stevieegee:
I enable MySQL 8.0 port 3306 or 3308 and MySQL 5.7 port 3307. Both work fine until I restart my Mac. Then with MySQL 8.0 is get the error - MySQL said: Host '172.18.0.1' is not allowed to connect to this MySQL server. MySQL 5.7 is alway fine just 8.0. I have deleted them and their volumes and recreated them in a different order but 8.0 always gives me the same error.

stevieegee:
The problem with MySQL 8.0 seems to be due to importing multiple database. If I  mysql -u root -p --host='127.0.0.1' --port=3306 < all_db.sql , it imports all the dbs and works fine, but if I restart the container I get the error.

mattstauffer:
@stevieegee that's very strange. Are you still seeing this error? Thanks!

mabadir:
@mattstauffer I am seeing that error as well. This happens usually while Docker is spinning the container. Once it's fully operational the error goes away.

mattstauffer:
OK, that makes sense. Database doesn't connect while it's restarting? Is anyone seeing it any other times?

MarceauKa:
Got the same error after a takeout + docker reinstallation. When I sshed into the container, I can't execute a simple `mysql -u root -p`. So locally and remotly I am unable to use my container

josecanhelp:
This seems like an issue with the container service not having enough time to get to a healthy state before trying to access the service. In the node port of takeout, we will be implementing a listener to check the health state before allowing interaction with the service.

PR #259, opened 2021-11-12
MySQL on M1

tonysm:
### Changed

- Changes the MySQL service to use `mysql/mysql-server`

### Fixed

- Fixes the `latest` tag resolution to take into account M1 chips (`arm64` architecture) when deciding which tag to use
- Fixes not being able to connect to the container from host with root due to missing `MYSQL_ROOT_HOST` env var (https://github.com/tighten/takeout/issues/190)

---

<details>

<summary>Screenshot</summary>

image

</details>

### DIFF
--- a/app/Shell/DockerTags.php
+++ b/app/Shell/DockerTags.php
@@ -71,12 +71,12 @@ class DockerTags
      */
     protected function armSupportedImagesOnlyFilter()
     {
-        return function ($results) {
+        return function ($tags) {
             $platform = $this->platform();
 
             // We need to take into account if the M1 chip is supported by the tag.
-            return $results->filter(function ($results) use ($platform) {
-                return collect($results['images'])
+            return $tags->filter(function ($tag) use ($platform) {
+                return collect($tag['images'])
                     ->pluck('architecture')
                     ->contains($platform);
             });
@@ -90,9 +90,9 @@ class DockerTags
      */
     protected function nonArmOnlySupportImagesFilter()
     {
-        return function ($results) {
-            return $results->filter(function ($results) {
-                $supportedArchitectures = collect($results['images'])
+        return function ($tags) {
+            return $tags->filter(function ($tag) {
+                $supportedArchitectures = collect($tag['images'])
                     ->pluck('architecture')
                     ->unique()
                     ->values();
--- a/app/Shell/DockerTags.php
+++ b/app/Shell/DockerTags.php
@@ -74,7 +74,8 @@ class DockerTags
         return function ($tags) {
             $platform = $this->platform();
 
-            // We need to take into account if the M1 chip is supported by the tag.
+            // We only want tags that specify arm64 in the supported architecture.
+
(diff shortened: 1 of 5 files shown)


tonysm:
Tested it on an M1 Mac and my Ubuntu laptop. Both work.

tonysm:
Hold on merging this. Noticed the container is not starting on Linux. Taking a look

tonysm:
I had to ditch my previous MySQL volume. But, after that, like everything is working.

app/Shell/DockerTags.php
```diff
@@ -61,6 +64,54 @@ public function getTags(): Collection
...
+        return function ($tags) {
+            $platform = $this->platform();
+
+            // We only want tags that specify arm64 in the supported architecture.
+
+            return $tags->filter(function ($tag) use ($platform) {
+                return collect($tag['images'])
+                    ->pluck('architecture')
+                    ->contains($platform);
+            });
```
mattstauffer, line 83:
```suggestion
            // Filter to only tags that list "arm64" in the supported architecture
            return $tags->filter(function ($tag) use ($platform) {
                return collect($tag['images'])
                    ->pluck('architecture')
                    ->contains($this->platform());
            });
```

app/Shell/DockerTags.php
```diff
@@ -61,6 +64,54 @@ public function getTags(): Collection
         return $sortedTags->values()->filter();
     }
 
+    /**
+     * Filter that ensures images that do not support arm architecture are filtered out.
```
mattstauffer, line 68:
```suggestion
     * Return a function intended to filter tags, ensuring images that do not support arm architecture are filtered out.
```

app/Shell/DockerTags.php
```diff
@@ -61,6 +64,54 @@ public function getTags(): Collection
...
+            return $tags->filter(function ($tag) use ($platform) {
+                return collect($tag['images'])
+                    ->pluck('architecture')
+                    ->contains($platform);
+            });
+        };
+    }
+
+    /**
+     * Filter that ensures are arm-only images are filtered out.
```
mattstauffer, line 88:
```suggestion
     * Return a function intended to filter tags, that ensures are arm-only images are filtered out.
```

app/Shell/DockerTags.php
mattstauffer: Also... we're directly calling "arm64" in the next function, but using `$this->platform()` here to refer to `arm64`. Obviously in the next function `arm64` won't actually be the result of `platform()`, but I'm still inclined just to use a string in both cases. So maybe instead of `->contains($this->platform())`, let's just do `->contains('arm64')`?

mattstauffer (changes requested):
Great work! A few notes

app/Shell/DockerTags.php
tonysm: Done!

mattstauffer:
WOO @tonysm !!

PR #263, opened 2021-11-30
Add `phpunit.xml` to the `.gitignore` file

markwalet:
I see the project already makes use of the `phpunit.xml.dist` file instead of the normal `phpunit.xml` file. This makes it possible to make local-only changes to your testsuite.

To fully support this, it is common to ignore the phpunit.xml file in projects.

The PR's will not get smaller than this one ;p

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /.vagrant
 .phpunit.result.cache
 /storage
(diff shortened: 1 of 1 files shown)


mattstauffer:
Haha, thanks @markwalet!

PR #265, opened 2021-12-17
Set Minio domain to allow dns buckets

### LINKED ISSUE #264: Feature Request - host-based buckets for Minio
minio has included support for host-based (subdomain) buckets for several years, to use it, it's necessary to pass a flag to minio specifying the domain.  AWS will be (at some point) deprecating path-based buckets, so it would be good to at least optionally support this before then.

From a minio github issue back in 2017:
> This feature is on only when server is started with --address <fqdn>

From the docs:
> By default, MinIO supports path-style requests that are of the format http://mydomain.com/bucket/object. MINIO_DOMAIN environment variable is used to enable virtual-host-style requests. If the request Host header matches with (.+).mydomain.com then the matched pattern $1 is used as bucket and the path is used as object.
>    
>    Example: 
>    
>    ```
>    export MINIO_DOMAIN=mydomain.com
>    
>    minio server /data
>    ```

I believe this means if we optionally collect a domain name (ie: minio.test, minio.localdomain, etc) and pass that into the docker container as an environment variable, dns buckets should work, which would future proof for whenever amazon eventually drops path based bucket support, and allow use of current tools that (kind of stupidly, imho) don't support path-based buckets.

willvincent:
Closes #264.

However, it would probably be preferable for the domain default to be null and then to not include `MINIO_DOMAIN` in the run template at all, it wasn't immediately clear to me how that would best be accomplished, so I instead opted for a sensible default of localhost.

### DIFF
--- a/app/Services/Minio.php
+++ b/app/Services/Minio.php
@@ -24,6 +24,11 @@ class Minio extends BaseService
             'default' => 9001,
         ],
         [
+            'shortname' => 'domain',
+            'prompt' => 'What domain will Minio be accessible at (optional, to allow dns buckets)?',
+            'default' => 'localhost',
+        ],
+        [
             'shortname' => 'root_user',
             'prompt' => 'What will the root user name for Minio be?',
             'default' => 'minioadmin',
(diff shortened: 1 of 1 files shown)


mattstauffer:
@willvincent Hm, I *think* we have another one somewhere with a conditional flag? Let me try to hunt it down

mattstauffer:
https://github.com/tighten/takeout/blob/main/app/Services/Sftp.php#L66

You can see a conditional modification of the `dockerRunTemplate` here, in `Sftp`.

willvincent:
Ahh sweet, I'll update accordingly, thanks

mattstauffer:
Thanks @willvincent!

willvincent:
My pleasure, sir.

PR #274, opened 2022-02-24
Upgrade Laravel Zero to v9

bandgeekndb:
Upgrades the Laravel Zero package to v9, which drops PHP7 support. I bumped the associated dependencies to ensure we got the correct minimum versions and also bumped the tlint package to v6 to satisfy the requirements.

I was successfully able to build and run Takeout using PHP 8.1.3 after performing this upgrade.


mattstauffer:
Fantastic, thanks @bandgeekndb!

scrummitch:
Congrats 🎆 great package

PR #278, opened 2022-03-04
Commit composer.lock and update build

bandgeekndb:
This fixes the issues introduced by PR #274. I forgot to commit the composer.lock file and I believe that's why we're seeing issues with the 2.0.0 build. I've included a v2.0.1 build that works, so this should just need a tag.


mattstauffer:
Thanks @bandgeekndb!

PR #286, opened 2022-04-15
Improve Start and Stop DX

### LINKED ISSUE #284: enable and start use different parameters
Start needs a container image but enable needs a service name - any chance these can be `consistantified*`?

 takeout start mysql8.0

 ERR  Docker container mysql8.0 not found.

 takeout start ce960ac00aad
 OUT  ce960ac00aad

takeout disable postgres
 OUT  3dbce21477fd
Disabling Service...: ✔

`* made consistent ;)`

tonysm:
### Changed

- Adds the ability to start and stop containers by service name. Currently, to start/stop specific containers we need to pass the container ID. With this change, we get the ability to also pass the service name of such containers. This makes the start/stop commands behave closer to enable/disable.

---

Before:

```bash
takeout start 123123123
```

Now:
```bash
takeout start mysql
```

resolves #284

### DIFF
--- a/app/Commands/DisableCommand.php
+++ b/app/Commands/DisableCommand.php
@@ -75,7 +75,7 @@ class DisableCommand extends Command
                 $serviceContainerId = $serviceMatches->flip()->first();
                 break;
             default: // > 1
-                $serviceContainerId = $this->selectMenu($disableableServices ?? $this->disableableServices);
+                $serviceContainerId = $this->selectMenu($this->disableableServices);
 
                 if (! $serviceContainerId) {
                     return;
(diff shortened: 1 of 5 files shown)


app/Commands/DisableCommand.php
```diff
@@ -75,7 +75,7 @@ public function disableByServiceName(string $service): void
                 $serviceContainerId = $serviceMatches->flip()->first();
                 break;
             default: // > 1
-                $serviceContainerId = $this->selectMenu($disableableServices ?? $this->disableableServices);
+                $serviceContainerId = $this->selectMenu($this->disableableServices);
```
tonysm, line 78: There was no `$disableableServices` variable in scope here.

vblinden:
I was about to build this, but saw that this pull request was already made. @mattstauffer do you have time to take another look at this?

PR #288, opened 2022-05-26
Add prompts for security on OpenSearch and ElasticSearch services

datashaman:
Using the search services without security is not simply a case of changing client config, the services need to be configured as well.

This makes using Opensearch and Elasticsearch without security or SSL possible. The defaults are set to secure.

Note that the opensearch change enables HTTP access, which is less troublesome than HTTPS.

The Elasticsearch change enables access without having to provide username and password like:
```
curl http://elastic:changeme@127.0.0.1:9200
```

HTTPS access on Elasticsearch is only available to paid subscriptions and requires a license key, which I dont think this tool would use.


mattstauffer:
Thanks @datashaman!

PR #289, opened 2022-05-26
Add builds folder to .gitignore

datashaman:
I notice the builds folder with the takeout binary is not ignored, someone will commit that at some point by mistake. This ignores the folder.


mattstauffer:
The `builds` folder *is* committed; that's the actual binary that's being distributed.

PR #294, opened 2022-07-03
Allow passing through additional options after the enable command

mattstauffer:
E.g.

```bash
takeout enable meilisearch -- --meilisearch-specific-flag-here --other-flag=great
```

I *think* it works correctly, but some failing tests I'm looking into, and I haven't actually tested it the whole way through to actually assuring the flags actually were passed the whole way to Docker the way I want.

### DIFF
--- a/app/Commands/EnableCommand.php
+++ b/app/Commands/EnableCommand.php
@@ -26,13 +26,19 @@ class EnableCommand extends Command
         $this->services = $services;
         $this->initializeCommand();
 
-        $services = $this->argument('serviceNames');
+        $services = $this->removePassthroughOptions($this->argument('serviceNames'));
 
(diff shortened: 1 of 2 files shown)


mattstauffer:
OK, it's incomplete in part because I only am processing pass through flags that start with two dashes, so i need to write a test and fix it to also work for this:

```bash
takeout enable meilisearch -- -e MEILI_MASTER_KEY='MY_KEY_HERE'
```

mattstauffer:
Fixed. Think this is ready to be tested?

mattstauffer:
Fixed that.

mattstauffer:
The failing tests are a PHP version issue that we're handling elsewhere.

PR #298, opened 2022-08-26
Changed the MSSQL image for one that works on M1

### LINKED ISSUE #293: SQL Server on Mac M1 fails to start
Fails with the message: ` ERR  WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested`

If it helps, the following command works `docker run -e "ACCEPT_EULA=1" -e "MSSQL_SA_PASSWORD=MyPass@word" -e "MSSQL_PID=Developer" -e "MSSQL_USER=SA" -p 1433:1433 -d --name=sql mcr.microsoft.com/azure-sql-edge`

Taken from https://medium.com/geekculture/docker-express-running-a-local-sql-server-on-your-m1-mac-8bbc22c49dc9

tonysm:
### Changed

- Changes the MSSQL image from `mssql/server` to `azure-sql-edge` which seems to be one that works on both M1 and Intel-based machines.

---

Note: We have only tested that when using the new image, the container (service) is able to run on M1 machines, we didn't fully tested if this image is a full replacement of the previous one. [The issue](https://github.com/tighten/takeout/issues/293) of where this change originated from mentions [this article](https://medium.com/geekculture/docker-express-running-a-local-sql-server-on-your-m1-mac-8bbc22c49dc9) which seems to imply it's a replacement.

resolves #293

### DIFF
--- a/app/Services/MsSql.php
+++ b/app/Services/MsSql.php
@@ -9,7 +9,7 @@ class MsSql extends BaseService
     protected static $category = Category::DATABASE;
 
     protected $organization = 'mcr.microsoft.com';
-    protected $imageName = 'mssql/server';
+    protected $imageName = 'azure-sql-edge';
     protected $dockerTagsClass = MicrosoftDockerTags::class;
     protected $defaultPort = 1433;
     protected $prompts = [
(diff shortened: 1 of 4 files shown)


mattstauffer:
@tonysm I'm sorry for taking so long. This looks good; would you mind resolving the conflicts, making sure it still works, and then you can merge this and tag a release? Thanks!

tonysm:
@mattstauffer merged `main` here and resolved the conflict. Tested again and the problem described in #293 still happens in `main`. Tested this fix and looks like it still works!

Issue #301, opened 2022-11-09
Shareable Volume Between Container and Host


devsquad-alvaro-meireles:
Hey, folks!

I want to know how I can use Takeout MySQL as a way to create a volume to share data between the container and the host. In my current dev environment I use something like that:

![image](https://user-images.githubusercontent.com/98560934/200902647-6d80ab0b-477a-4dac-9f05-0505cc2f2f1c.png)

... So inside the container I have the: "/var/download" that is my "Download" folder, so I'm able to, for example, restore a database using a dump.

How I can have this same behavior using Takeout?

josecanhelp:
It sounds like you want to mount a local path into the container so that a dump into that directory would be saved on your host system. Takeout does not offer a way to mount local paths. It only creates a default docker volume so that database data does not get lost when destroying a container.

By design, Takeout is opinionated in order to simplify the creation and destruction of service containers for developers that don't want to deal with the complexity of using vanilla docker commands.

What you are trying to do can be done with a little workaround. For example, you could grab [the command](https://github.com/tighten/takeout/blob/7ae40aa900db046cab3fae25307169c8a7851331/app/Services/MySql.php#L25)  that Takeout uses to enable your mysql container and add another `-v` flag. An example command might look like this:

```
docker run -p 3306:3306 \
        -e MYSQL_ROOT_PASSWORD=secret \
        -e MYSQL_ALLOW_EMPTY_PASSWORD=true \
        -e MYSQL_ROOT_HOST=% \
        -v mysql_data:/var/lib/mysql \
        -v ~/Downloads:/var/downloads \
        mysql/mysql-server:8.0.31-amd64 --default-authentication-plugin=mysql_native_password
```
Just note that creating a container manually like this means that Takeout has no knowledge of the container and cannot help enabling/disabling it. You would need to run vanilla docker commands to do that.

`docker ps` to list containers
`docker stop CONTAINER_ID` to stop containers
`docker start CONTAINER_ID` to start containers
`docker rm CONTAINER_ID` to remove containers
`docker volume ls` to list docker volumes
`docker volume rm VOLUME_NAME` to remove volumes

josecanhelp:
To add to this, I noticed that Takeout is now allowing passthrough parameters. However, there seems to be a small bug with that feature at the moment. I created a PR #302 to resolve this. Once we can get a fix merged in, it should be relatively simple to add a custom flag when enabling a container like so:

`takeout enable mysql -- -v ~/Downloads:/var/downloads`

[...]

PR #302, opened 2022-11-09
Fix ordering issue with passthrough parameters

### LINKED ISSUE #299: Docker run command parameter order
@mattstauffer I believe that I have identified a bug with the new implementation of #218.  Currently the implementation appends the command to the end of the `$dockerRunTemplate` for each service.  
```php
protected $dockerRunTemplate = '-p "${:port}":1433 \
        -e ACCEPT_EULA=Y \
        -e "TZ=America/Chicago" \
        -e SA_PASSWORD="${:sa_password}" \
        -v "${:volume}":/var/opt/mssql \
        "${:organization}"/"${:image_name}":"${:tag}"';
```
So say I was trying to run the following command:
` takeout enable mssql --default -- -e '"TZ=America/Chicago"' `
I would get the following output if I was dumping the console commands.
```php
docker run -d --name "${:container_name}" --network=takeout --network-alias="${:alias}" --label com.tighten.takeout.Full_Alias=mssql2019-GA-ubuntu-16.04 --network-alias="mssql/server" --label=com.tighten.takeout.Base_Alias=mssql/s
erver -p "${:port}":1433 \n
        -e ACCEPT_EULA=Y \n
        -e SA_PASSWORD="${:sa_password}" \n
        -v "${:volume}":/var/opt/mssql \n
        "${:organization}"/"${:image_name}":"${:tag}" -e "TZ=America/Chicago"
```
The command will run without error; however, the container will not start and will give the following error.
```
/opt/mssql/bin/permissions_check.sh: line 59: exec: -e: invalid option
exec: usage: exec [-cl] [-a name] [command [arguments ...]] [redirection ...]
SQL Server 2019 will run as non-root by default.
This container is running as user mssql.
Your master database file is owned by mssql.
To learn more visit https://go.microsoft.com/fwlink/?linkid=2099216.
```
This is because per the Docker docs  for the run command (https://docs.docker.com/engine/reference/run/) the options should come before the image.
`$ docker run [OPTIONS] IMAGE[:TAG|@DIGEST] [COMMAND] [ARG...]`

I would like to suggest the following implementation to fix this issue.  

```php
 protected $dockerRunTemplate = '-p "${:port}":1433 \
        -e ACCEPT_EULA=Y \
        -e SA_PASSWORD="${:sa_password}" \
        -v "${:volume}":/var/opt/mssql \
>>>     $optional_parameters \
        "${:organization}"/"${:image_name}":"${:tag}"';
```
All of the Service templates would need to be updated and code added to inject the new variable into the template.

I have verified this works by manually updating the template for mssql to the following and ran and the container started successfully and the timezone value was updated for the container.

```php
protected $dockerRunTemplate = '-p "${:port}":1433 \
        -e ACCEPT_EULA=Y \
        -e "TZ=America/Chicago" \
        -e SA_PASSWORD="${:sa_password}" \
        -v "${:volume}":/var/opt/mssql \
        "${:organization}"/"${:image_name}":"${:tag}"';
```
Let me know if you would like for me to work on a PR to address this issue.

Regards,
Kevin

josecanhelp:
Passthrough parameters were being added to the end of the run commands, after the image name and tag. However, those parameters should be passed through to the command before that. The simplest solution, in my opinion, is to reorder those strings at the base command enable function. I tested this with passing custom ports as well as custom volumes.

This PR would resolve #299 as well as #301.

### DIFF
--- a/app/Services/BaseService.php
+++ b/app/Services/BaseService.php
@@ -79,7 +79,7 @@ abstract class BaseService
 
         try {
             $this->docker->bootContainer(
-                $this->sanitizeDockerRunTemplate($this->dockerRunTemplate) . $this->buildPassthroughOptionsString($passthroughOptions),
+                trim("".join(" ", [$this->buildPassthroughOptionsString($passthroughOptions), $this->sanitizeDockerRunTemplate($this->dockerRunTemplate)])),
                 $this->buildParameters()
             );
 


devsquad-alvaro-meireles:
Hey, @mattstauffer @tonysm  ! Can you review that please?

tonysm:
I'll check it tomorrow (it looks good, but the build is not passing)

devsquad-alvaro-meireles (approved):
🚀 🔥

ajmeireles (approved):

tonysm (approved):

tonysm:
@josecanhelp @devsquad-alvaro-meireles one question: I think the original feature allowed passing arguments to the container process, but with this change, the passthrough feature is now for docker run options, or am I reading this wrong? 🤔 

before:
```bash
docker run image {container-args}
```

after:
```bash
docker run {docker-run-options} image
```

josecanhelp:
@tonysm that seems right. I thought the pass through was for docker run commands, not container processes.

tonysm:
@josecanhelp maybe we should offer a new option for the run command, something like:

```bash
takeout enable mysql --run="--restart unless-stopped -e 'LOREM=IPSUM'" --  -hsome.mysql.host -usome-mysql-user -p
```

where it would generate a command like:

```bash
docker run {other-options} --restart unless-stopped -e 'LOREM=IPSUM' mysql -hsome.mysql.host -usome-mysql-user -p
```

ajmeireles:
> @josecanhelp @devsquad-alvaro-meireles one question: I think the original feature allowed passing arguments to the container process, but with this change, the passthrough feature is now for docker run options, or am I reading this wrong? thinking
> 
> before:
> 
> ```shell
> docker run image {container-args}
> ```
> 
> after:
> 
> ```shell
> docker run {docker-run-options} image
> ```

@tonysm The main idea that I've asked in #301 was the ability to set a custom volume to be shareable with the container, so we can easily share a database dump to the MySQL container

tonysm:
We could offer the extra `--volume` flag in the enable command, but my issue with that is that we would eventually be adding more and more docker run options to the list of options, and the docker run command _"has more options than any other docker command"_ ([stated by the docs](https://docs.docker.com/engine/reference/run/)).

So, I'd be more inclined to offer a single option for power users to pass any docker run options to the docker run command underneath. Thoughts?

ajmeireles:
> So, I'd be more inclined to offer a single option for power users to pass any docker run options to the docker run command underneath. Thoughts?

Sounds good! :rocket:

tonysm:
Alright, so I'll revert this change and introduce a new `--run` option to the enable command to pass docker run options.

tonysm:
The [v2.2.0](https://github.com/tighten/takeout/releases/tag/v2.2.0) was tagged with the new `--run=` option. \cc @josecanhelp @ajmeireles

PR #304, opened 2022-11-18
Adds a `--run` option to the enable command

tonysm:
### Changed

- Reverts the passthrough options to be container arguments instead of docker run options (https://github.com/tighten/takeout/pull/302)

### Added

- Adds a new `--run` option to the enable command which can be used to pass any extra docker run option

---

This is how we can use this:

```bash
takeout enable mysql --run="--restart unless-stopped -e 'LOREM=IPSUM'" -- -h127.0.0.1 -usome-user
```

This will be translated to a docker command like this:

```bash
docker run --restart unless-stopped -e 'LOREM=IPSUM' {service-options} mysql -h127.0.0.1 -usome-user
```

where `{service-options}` depends on the options defined in the service class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ You may specify extra arguments to the container after a `--` sepatator:
 takeout enable mysql -- -hsome.mysql.host -usome-user
 ```
 
-Notice that these are options for the container (arguments), not extra docker run options (see below).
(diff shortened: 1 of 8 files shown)


driftingly (approved):
This looks really good @tonysm!
We'll need to document this and how to pass through additional options `--` when we tag the next release.

Issue #296, opened 2022-07-28
PHP error in phar


MarceauKa:
Hi, after removing then reinstalling takeout I got this error when I try to run `takeout`

```php
Parse error: syntax error, unexpected token ")" in phar:///Users/username/.composer/vendor/tightenco/takeout/builds/takeout/vendor/symfony/finder/Finder.php on line 588
```

Takeout: 2.1
Composer: 2.3.5
PHP: 8.0.21

mattstauffer:
Thanks! Closed in 2.1.1

subdesign:
Hi, I have the same error, but I have 2.2.0 installed! 
PHP 8.0.19
Comp: 2.3.5

<img width="1351" alt="Screenshot 2022-12-02 at 11 15 47" src="https://user-images.githubusercontent.com/691695/205270187-c03dbf81-04f6-4c28-98dd-ce38ea253efe.png">

therobfonz:
Also seeing this same error on 2.3.0 and PHP 8.0.26. 

![Screenshot 2023-01-03 at 6 05 02 AM](https://user-images.githubusercontent.com/35386780/210345386-7efb4147-8e48-4e35-a016-cafb77e302fa.png)

tonysm:
Taking a look at this...

tonysm:
So, packagist says we support PHP 8.0. To do that, we need to be on that version when developing, or at least at the time we're building the phar file and also when installing the Composer dependencies. Installing the Composer dependencies from a higher PHP version may get dependencies that are not compatible with PHP 8.0.

tonysm:
Just tagged a new patch release [v2.3.1](https://github.com/tighten/takeout/releases/tag/v2.3.1) with the fix for this.

PR #308, opened 2023-01-03
Ensure Composer dependencies are compatible with PHP 8.0

tonysm:
### Fixed

- Fixes `phar` file not compatible with PHP 8.0 when pulling the Composer dependencies and building from a higher PHP version

---

issues #296

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -47,7 +47,7 @@ jobs:
                   coverage: none
 
             - name: Install dependencies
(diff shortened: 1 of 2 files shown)


driftingly (approved):
Setting `platform-check` to `true` checks php and extensions. 
I was reading a bit about it here:
https://php.watch/articles/composer-platform-check

> Composer v2 now creates a new vendor/composer/platform_check.php file that immediately terminates rest of the application if the current server platform does not match the original environment it was created on.

It goes on to show error messages
> Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 7.3.0" and "< 7.4.0". You are running 7.4.5.

So it looks like it will work as-is now.

PR #311, opened 2023-02-06
Update MailDev image tag, ports and template

thomasakarlsen:
### Changed

- Updated default image tag for maildev
- Changed ports used by maildev image

---
MailDev has released a new major version that has a breaking change in the ports exposed by the docker image. With this new major version they have also introduced ARM image support, helpful for us running on newer macs.


tonysm (approved):
Looks good! 👍🏼 Thanks!

PR #313, opened 2023-03-14
EventStoredB: use latest version, pass in params for local dev

morrislaptop:
Since #154 , EventStore has improved the local DX. This PR reverts to use the latest version - which to my benefit will auto-select the arm version on my mac, love it!


tonysm (approved):
Looks good. I was able to run the container from my M1 mac with no issue. Thanks! 👍🏼

PR #317, opened 2023-04-26
Add traefik service

viicslen:
Adds [traefik](https://doc.traefik.io/traefik/) reverse proxy to the list of available services

### DIFF
--- a/app/Services/Traefik.php
+++ b/app/Services/Traefik.php
@@ -1,0 +1,86 @@
+<?php
+
+namespace App\Services;
+
+use App\Shell\Docker;
+use App\Shell\Environment;
+use App\Shell\Shell;
+use Exception;
+
+class Traefik extends BaseService
+{
+    protected static $category = Category::TOOLS;
+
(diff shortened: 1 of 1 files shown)


mattstauffer:
In one of your comments you wrote this: 

```
// Cannot use $_SERVER superglobal since that's empty during UnitUnishTestCase;
```

What do you mean by `UnitUnishTestCase`?

mattstauffer:
Ha! OK, I'll remove.

viicslen:
> Extracted and re-wrote here: [6eda6b5](https://github.com/tighten/takeout/commit/6eda6b577134041657f33d7e61feaba0ceb44f14)

Yeah, that looks much better.

BTW, I'm not too sure but I think the Windows directory separator should also be included when trimming the trailing slash in the following line:
https://github.com/tighten/takeout/blob/6eda6b577134041657f33d7e61feaba0ceb44f14/app/Shell/Environment.php#L77

mattstauffer:
whoops, thanks! i missed that.

PR #319, opened 2023-06-06
[Fix][Traefik] Update prompt name key when checking for empty config directory

viicslen:


tonysm (approved):
Thanks for the fix! Looks like we're not properly setting the default value on line 59. I'll fix it in a follow up PR.

Issue #324, opened 2023-08-31
Add Prompt interface to stock enable command


njames:
Add the new fancy Prompts interface to the enable command.

njames:
So takeout has 

```json
        "platform": {
            "php": "8.0.2"
        }
```
even thought it has `php ^8.1`

So `composer require laravel/prompts` fails as 

```bash
Root composer.json requires php ^8.1 but your php version (8.0.2; overridden via config.platform, actual: 8.2.9) does not satisfy that requirement.
```

Can we bump config.platform to at least 8.1 please?

njames:
or just remove it 

I have jiggled the composer.json here 
https://github.com/squarecloudlabs/takeout/blob/add-prompts-to-enable/composer.json

Issue #282, opened 2022-03-10
Menu too tall for some screens


mattstauffer:
<img width="823" alt="image" src="https://user-images.githubusercontent.com/151829/157705193-eee4a748-6d9a-41d4-9b73-102839340299.png">

Let's look into whether the tool we're using to generate these has multi page options?

tonysm:
@mattstauffer I've looked into the Laravel Console Menu package ([source](https://github.com/nunomaduro/laravel-console-menu)) and the underlying `cli-menu` package from php-school ([source](https://github.com/php-school/cli-menu)) and looks like there's no option for pagination/scrolling.

We could, technically, limit the number of items displayed in the menu and add an extra "Show More" option or "Next" that would re-render the menu with more options (then showing "Prev" and "Next" options on subsequent pages). Basically implementing pagination manually on top of this menu package.

I personally find this kind of menu a bit limiting (it makes testing more complicated and I'm not very fond of the looks of it), wouldn't reimplementing this in a regular "choices" menu be a better solution? We could even have interactive autocompletion, like:

https://user-images.githubusercontent.com/1178621/164785821-2fd85e61-cadc-4d05-b540-0da7063bf6b1.mp4

chrisdicarlo:
> @mattstauffer I've looked into the Laravel Console Menu package ([source](https://github.com/nunomaduro/laravel-console-menu)) and the underlying `cli-menu` package from php-school ([source](https://github.com/php-school/cli-menu)) and looks like there's no option for pagination/scrolling.
> 
> We could, technically, limit the number of items displayed in the menu and add an extra "Show More" option or "Next" that would re-render the menu with more options (then showing "Prev" and "Next" options on subsequent pages). Basically implementing pagination manually on top of this menu package.
> 
> I personally find this kind of menu a bit limiting (it makes testing more complicated and I'm not very fond of the looks of it), wouldn't reimplementing this in a regular "choices" menu be a better solution? We could even have interactive autocompletion, like:
> 
>  Screen.Recording.2022-04-22.at.17.00.28.mp4

I like that option!

khanalpride:
I have implemented this feature as below:

We can divide the service types and containers for each types into two display like this?

<img width="706" alt="takeout-menu1" src="https://user-images.githubusercontent.com/6267947/204975030-0373f787-4f32-41a7-b22d-202d23bad593.png">


After "Database" is selected in the first menu? It shows like this:
[...]

PR #328, opened 2023-10-27
Use Prompts

gcavanunez:
### Changed

- Drops the `nunomaduro/laravel-console-menu` dependency together with the `ext-posix` and `ext-pcntl` extensions requirement (`posix` was only used by the menu, not sure why we had `pcntl` listed since we're not using process signals)
- Adds Laravel Prompts and replaces our outputs in most places (some places, like the process output, were kinda weird when trying to use prompts... we also tried Termwind, but we figured it wasn't worth it for now... at least until we figure out a better TUI for this). We're using prompts for the enable menu, but also for the "prompts" (our own additional questions were called that), which makes a nice experience. We have also converted the functions to use Prompts validations. 
- We're also changing how we detect if a port is available. Previously, we were relying on `netstat` being installed. Instead, we will switch to attempting to do a `fsockopen()` on the port. If it returns a resource, it means something is running on that port, so it's NOT available. Otherwise, it returns false, which means nothing is running on it so the port is available. We're using the `TAKEOUT_CONTAINER=1` environment variable (already present in Takeout's Docker image) to detect if we should check `localhost` or `host.docker.internal` host instead. We're also updating the docs to add the `--add-host` option needed for this to work on Linux (it already works without it on Windows and Mac).
- The `logs` command now accepts the name of the service... and when passed nothing, it will prompt the user with the options of running Takeout containers

---

closes https://github.com/tighten/takeout/issues/324
closes https://github.com/tighten/takeout/issues/282


gcavanunez:
After upgrading to ` "laravel/prompts": "^0.1.13" ` the error with the console being prompted during tests go away 

This upgrade does require ` "laravel-zero/framework": "^10.1.1" `

mattstauffer:
@gcavanunez I'm super interested in getting this merged. The install box has gotten out of hand! Do you need any help to get this moved forward?

gcavanunez:
@mattstauffer 100%!! Since this guy depends on bumping up to the latest version of laravel-zero, taking a look at this PR could trickle this one out 
- https://github.com/tighten/takeout/pull/339

app/Commands/StartCommand.php
```diff
@@ -64,8 +64,8 @@ public function startableContainers(): Collection
     public function startByServiceNameOrContainerId(string $service, Collection $startableContainers): void
     {
         $containersByServiceName = $startableContainers
-            ->filter(function ($serviceName) use ($service) {
-                return Str::startsWith($serviceName, $service);
+            ->filter(function ($serviceName, $key) use ($service) {
+                return Str::startsWith($serviceName, $service) || $key === $service;
```
tonysm, line 68: nice!

app/Services/BaseService.php
```diff
@@ -160,29 +160,30 @@ protected function prompts(): void
...
-                app('console')->error("Port {$this->promptResponses[$prompt['shortname']]} is already in use. Please select a different port.");
-                $this->askQuestion($prompt);
-            }
-
-            $items[] = $prompt;
+        $questions = array_merge($this->defaultPrompts, $this->prompts);
+
+        foreach ($questions as $prompt) {
+            $items[] = match (true) {
+                Str::contains($prompt['shortname'], 'port') => $this->askQuestion($prompt, $this->useDefaults, validate: function (string $port) {
```
mattstauffer, line 167: Why are we moving to `Str::contains`? It makes it less clear what we're doing.

app/Shell/Environment.php
```diff
@@ -28,35 +26,26 @@ public function isWindowsOs(): bool
...
 
-    public function portIsAvailable($port): bool
+    public function portIsAvailable(int $port): bool
     {
-        // E.g. Win/Linux: 127.0.0.1:3306 , macOS: 127.0.0.1.3306
-        $portText = $this->isLinuxOs() ? "\:{$port}\s" : "\.{$port}\s";
+        // To check if the socket is available, we'll attempt to open a socket on the port.
+        // If we cannot open the socket, it means there's nothing running on it, so the
+        // port is available. If we are successful, that means it is already in use.
+
```
mattstauffer, line 34: Drop the line after the comment block.

Also, I like that you made the fun Laravel -style three comments lines.. but unfortunately, I think we need a bit more -- or a bit different -- explanation. What's missing here is context for anyone who doesn't understand sockets. The natural expectation is that if I try to do a thing on a port and it works, that means the port was available, and if it fails, that means the port wasn't available. But this is the opposite. I think a good comment would explain *that*, more than explaining exactly what the following few lines do.

app/Shell/Environment.php
```diff
@@ -28,35 +26,26 @@ public function isWindowsOs(): bool
...
-            | grep '{$portText}' | grep -v 'TIME_WAIT' | grep -v 'CLOSE_WAIT' | grep -v 'FIN_WAIT'");
+        fclose($socket);
 
-        // A successful netstat command means a port in use was found
-        return ! $process->isSuccessful();
+        return false;
     }
 
-    public function netstatCmd(): string
+    public function isTakeoutRunningOnDocker(): bool
```
mattstauffer, line 46: Maybe `isTakoutRunningInsideDocker`? Unless that's not a Docker-y way to say it. But `Inside` definitely makes it clearer to me; my first thought reading this was, duh, of course it is 😆

mattstauffer (changes requested):
Incredible work! This is so exciting! Just a few notes.

app/Services/BaseService.php
tonysm: This was kinda duplicated. For the default prompts we were checking `shortname === port` but for the custom prompts (the services might have other port prompts), we were checking `Str::contains(shortname, 'port')`, so I figured it would be better to merge the prompts and just check once with Str::contains, which handles both cases.

app/Commands/DisableCommand.php
```diff
@@ -25,100 +30,76 @@ public function handle(Docker $docker, Environment $environment)
...
             return [$container['container_id'] => str_replace('TO--', '', $container['names'])];
-        })->toArray();
+        });
     }
 
-    public function disableByServiceName(string $service): void
+    private function disableByServiceName(string $service, Collection $disableableServices): void
     {
-        $serviceMatches = collect($this->disableableServices)
+        $serviceMatches = collect($disableableServices)
```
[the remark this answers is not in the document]
tonysm: Oh, sorry, I was in the wrong branch locally.

app/Exceptions/DockerMissingException.php
```diff
@@ -2,20 +2,16 @@
...
     {
-        $console = app('console');
-        $shell = app(Shell::class);
-
-        $console->line('');
-        $console->line($shell->formatErrorMessage('Docker is not installed.'));
-        $console->line('');
-        $console->line($shell->formatErrorMessage('Please visit https://docs.docker.com/get-docker/'));
-        $console->line($shell->formatErrorMessage('for information on how to install Docker for your machine.'));
+        error('Docker is not installed');
```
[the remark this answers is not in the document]
tonysm: Nope, we missed that.

app/Shell/Environment.php
tonysm: I've changed the comments

tonysm:
@mattstauffer ready for another review

app/Shell/Environment.php
```diff
@@ -28,35 +26,25 @@ public function isWindowsOs(): bool
         return PHP_OS_FAMILY === 'Windows';
     }
 
-    public function portIsAvailable($port): bool
+    public function portIsAvailable(int $port): bool
     {
-        // E.g. Win/Linux: 127.0.0.1:3306 , macOS: 127.0.0.1.3306
-        $portText = $this->isLinuxOs() ? "\:{$port}\s" : "\.{$port}\s";
+        // To check if the port is available, we'll attempt to open a socket connection to it.
+        // Note that the logic here is flipped: successfully openning the socket connection
```
mattstauffer, line 32:
```suggestion
        // Note that the logic here is flipped: successfully opening the socket connection
```

mattstauffer (changes requested):
Great work! A single typo in a comment and then this PR is good to merge!

mattstauffer (approved):

PR #330, opened 2023-12-12
fix issue #329

### LINKED ISSUE #329: use of MySql 8.2 and 8.0.35 -  official mysql docker hub image
Currently MySql service [https://github.com/tighten/takeout/blob/main/app/Services/MySql.php](https://github.com/tighten/takeout/blob/main/app/Services/MySql.php) uses mysql docker hub image from [https://hub.docker.com/r/mysql/mysql-server/](https://hub.docker.com/r/mysql/mysql-server/) which for the last time is upgraded more then 1 year go and does not include newly released MySql `8.2`. Last version available is  there is`8.0.32`

There is official mysql docker image location [https://hub.docker.com/_/mysql](https://hub.docker.com/_/mysql) which is maintained by [the Docker Community and the MySQL Team](https://github.com/docker-library/mysql) and is updated regularly and having all latest version like `8.2` and `8.0.35`.

My suggestion is to change MySql service [https://github.com/tighten/takeout/blob/main/app/Services/MySql.php](https://github.com/tighten/takeout/blob/main/app/Services/MySql.php) to use this official image which is updated more regularly [https://hub.docker.com/_/mysql](https://hub.docker.com/_/mysql).

i will provide PR

bnaydenov:
Make use of offical mysql docker image which includes latest versions 8.2 and 8.0.35.

remove variable `$organization` will default to string `library` which is used as `$organization` when pull official images.

fix #329

### DIFF
--- a/app/Services/MySql.php
+++ b/app/Services/MySql.php
@@ -6,8 +6,7 @@ class MySql extends BaseService
 {
     protected static $category = Category::DATABASE;
 
-    protected $organization = 'mysql';
-    protected $imageName = 'mysql-server';
+    protected $imageName = 'mysql';
     protected $defaultPort = 3306;
     protected $prompts = [
         [
--- a/app/Services/MySql.php
+++ b/app/Services/MySql.php
@@ -7,7 +7,9 @@ class MySql extends BaseService
     protected static $category = Category::DATABASE;
 
     protected $imageName = 'mysql';
+
     protected $defaultPort = 3306;
+
     protected $prompts = [
         [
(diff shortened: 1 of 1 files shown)


bnaydenov:
Can somebody from maintainers take a look why this test `Tests\Feature\DockerTagsTest::it_sorts_the_versions_naturally` is falling

tonysm (approved):
Looks like MySQL has caught up with Arm images. Nice. Thanks!

tonysm:
Thanks!

PR #334, opened 2024-03-10
Adds Buggregator as a tool

tonysm:
### Added

- Adds [Buggregator](https://buggregator.dev/) as a tool.

---

Hat tip to @gcavanunez for pointing me to Buggregator.

Usage:

```
takeout enable buggregator
```

After answering the prompted questions (about where the web UI, the VarDumper server, the Monolog server port, and network alias), the service should run at localhost:8000 (unless you change the default port).

### VarDumper Server

To start sending `dump()` calls to Buggregator, you may set the following configs on the Laravel app:

```env
VAR_DUMPER_FORMAT=server
VAR_DUMPER_SERVER=127.0.0.1:9912
# VAR_DUMPER_SERVER=buggregator:9912 # If you're on Sail
```

### Ray

First, on the Laravel app, require the Ray PHP client:

```bash
composer require spatie/laravel-ray --dev
```

Next, set the following configs:

```env
RAY_HOST=ray@127.0.0.1
# RAY_HOST=ray@buggregator # If you're on Sail
RAY_PORT=8000
```

### SMTP Server

To send emails, you may set the following configs on the Laravel app:

```env
MAIL_MAILER=smtp
MAIL_HOST=127.0.0.1
# MAIL_HOST=buggregator # If you're on Sail
MAIL_PORT=1025
```

### Profiler

Buggregator ships with an [XHProf](https://www.php.net/manual/en/book.xhprof.php) profiler. To use it, first install the `xhprof` extension via _pecl_:

```bash
pecl install xhprof
```

Then add the `extension=xhprof.so` to the `php.ini`. On Sail, you can do that with:

```bash
echo "extension=xhprof.so" >> /etc/php/8.3/cli/php.ini
```

Then, install the package:

```bash
composer require --dev maantje/xhprof-buggregator-laravel
```

Then, configure the profiler envs:

```env
PROFILER_ENDPOINT=http://profiler@127.0.0.1:8000
# PROFILER_ENDPOINT=http://profiler@buggregator:8000 # If you're on Sail
PROFILER_APP_NAME="${APP_NAME}"
XHPROF_ENABLED=true
```

### Sentry

First, install the package following sentry's [docs](https://docs.sentry.io/platforms/php/guides/laravel/#install).

Then, configure the Sentry DNS on the `.env` of your Laravel app:

```env
SENTRY_LARAVEL_DSN=http://sentry@127.0.0.1:8000/1
# SENTRY_LARAVEL_DSN=http://sentry@buggregator:8000/1 # If you're on sail
```

These should serve as examples. There's also a nice [monolog service](https://docs.buggregator.dev/config/monolog.html), an [inspector](https://docs.buggregator.dev/config/inspector.html), and an HTTP dump server.

### Notes on Sail

If you're using Takeout in combination with Sail, you can use the `network_alias` configured as the hostname (which defaults to `buggregator`) instead of `127.0.0.1`. The Laravel app container must be added to the Takeout network. For more info, check out [Matt's blog](https://mattstauffer.com/blog/how-to-use-takeout-to-add-new-services-to-laravel-sail-and-save-ram/).

### DIFF
--- a/app/Services/Buggregator.php
+++ b/app/Services/Buggregator.php
@@ -29,12 +29,17 @@ class Buggregator extends BaseService
             'prompt' => 'What is the Monolog port?',
             'default' => '9913',
         ],
+        [
+            'shortname' => 'network_alias',
+            'prompt' => 'What network alias to you want to assign to this container? This alias can be used by other services on the same network.',
+            'default' => 'dump-server',
+        ],
     ];
 
     protected $dockerRunTemplate = '-p "${:port}":8000 \
         -p "${:smtp_port}":1025 \
         -p "${:var_dumper_port}":9912 \
         -p "${:monolog_port}":9913 \
-        --network-alias dump-server \
+        --network-alias "${:network_alias}" \
         "${:organization}"/"${:image_name}":"${:tag}"';
 }
--- a/app/Services/Buggregator.php
+++ b/app/Services/Buggregator.php
@@ -32,7 +32,7 @@ class Buggregator extends BaseService
         [
             'shortname' => 'network_alias',
             'prompt' => 'What network alias to you want to assign to this container? This alias can be used by other services on the same network.',
-            'default' => 'dump-server',
+            'default' => 'buggregator',
         ],
     ];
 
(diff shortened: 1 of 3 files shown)


mattstauffer:
Oh, this is hot.

PR #338, opened 2024-04-22
Add Timescale

riasvdv:
This adds the [Timescale](https://www.timescale.com/) service


tonysm:
These macOS tests are too slow. I don't think there's anything broken, so I'm gonna skip them and merge. Thanks!

Issue #337, opened 2024-04-08
Unable to install mongo


rognales:
My machine is MBP M1 pro and installed version
```console
php -v
PHP 8.1.27 (cli) (built: Dec 19 2023 20:35:55) (NTS)

composer global show tightenco/takeout | grep versions
Changed current directory to /Users/rognales/.composer
versions : * v2.6.0
```
When i issue `takeout enable mongo --default`, it will failed with this

```console
takeout enable mongo --default

Warning: Module "imap" is already loaded in Unknown on line 0
Downloading docker image...

 OUT  7.0.8-nanoserver-ltsc2022: Pulling from library/mongo
 ERR  no matching manifest for linux/arm64/v8 in the manifest list entries
Enabling mongo...

 ERR  Unable to find image 'mongo:7.0.8-nanoserver-ltsc2022' locally
 ERR  7.0.8-nanoserver-ltsc2022: Pulling from library/mongo
 ERR  docker: no matching manifest for linux/arm64/v8 in the manifest list entries.
 ERR  See 'docker run --help'.

Failed installing mongo  
```

if i try to pass custom image `arm64v8/mongo` when it prompted for tags, it still down the same image as above

1. Is M1 not supported? But I can download the image manually `docker pull arm64v8/mongo`.
2. What is the correct way to pass custom image?

rognales:
it is pulling this image https://hub.docker.com/layers/library/mongo/7.0.8-nanoserver-ltsc2022/images/sha256-5623c8cd57549465794442099ddda3237e5043a6d21af96128dcef4bae2d6f5f?context=explore

<img width="1302" alt="image" src="https://github.com/tighten/takeout/assets/28396330/51d93a49-886d-45f4-b28d-9793fd32b6b1">

Perhaps it wrongly detect which arm variant?

rognales:
I managed to get it to run if I passed `jammy` as tag

tonysm:
Hey, M1 Mac is supported. I noticed a similar issue with the MySQL service pulling the `8.4.0-oraclelinux8` instead of just `8.4.0`, for instance. That has to do with the tagging sort, I believe.

PR #341, opened 2024-05-04
Fix Images Version Sorting

tonysm:
### Changed

- Fixes the version sorting so we consider semver image tags higher than the other distro-specific or flavored image tag names

---

closes https://github.com/tighten/takeout/issues/337

<details>
<summary>Screenshots</summary>

typesense
sqs
solr
singlestore
redis
postgres
postgis
opensearch
neo4j
mysql
mssql
mongo
minio
memcached
meilisearch
mariadb
mailpit
maildev
influxdb
expose
eventstoredb
elasticsearch

</details>

### DIFF
--- a/tests/Feature/VersionComparatorTest.php
+++ b/tests/Feature/VersionComparatorTest.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Tests\Feature;
+
+use App\Shell\VersionComparator;
+use Tests\TestCase;
+
+class VersionComparatorTest extends TestCase
+{
+    /**
+     * @test
+     *
+     * @testWith [["latest", "8.0"], ["latest", "8.0"]]
+     *           [["8.0", "latest"], ["latest", "8.0"]]
+     *           [["5.0", "8.0", "latest"], ["latest", "8.0", "5.0"]]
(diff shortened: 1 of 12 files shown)


tests/Feature/VersionComparatorTest.php
```diff
@@ -0,0 +1,32 @@
...
+
+use App\Shell\VersionComparator;
+use Tests\TestCase;
+
+class VersionComparatorTest extends TestCase
+{
+    /**
+     * @test
+     *
+     * @testWith [["latest", "8.0"], ["latest", "8.0"]]
```
gcavanunez, line 13: Nice!

gcavanunez (approved):
This is awesome! ran it locally, lists were sorting nicely, I feel like this really opens things up to turning this field into a combox/search select terminal prompt

jakebathman (approved):
This all looks good, nice work getting it more simplified and using that new comparator class. I haven't run this locally but if Guillermo has then that's great.

tonysm:
I've also tested this on an M1 and it works. One thing I want to eventually handle is that it should consider `8.0` as higher than `8.0.45`, but that can be enhanced later.

PR #115, opened 2020-09-12
Allow empty MySQL password

thecrypticace:
fixes #96

PR #343, opened 2024-05-29
Feat: Null password support for Mariadb

nikspyratos:
Heyo!

Took the logic from #115 and ported it for MariaDB. Spun up the container and I could connect with an empty password, so guessing that's all that needs to be done for testing.


nikspyratos:
Hmm, followed the existing code used, so for consistency I'd say keep it. Alternatively the entire repo probably needs a linting run at least once to prevent similar occurrences?

PR #344, opened 2024-06-14
Adds pgvector

### LINKED ISSUE #342: Add a pgvector service
I was playing with AI embedding last week and I think it would be cool to have a service dedicated to `pgvector`.

I used the [ankane/pgvector](https://hub.docker.com/r/ankane/pgvector) image, but there's also [pgvector/pgvector](https://hub.docker.com/r/pgvector/pgvector).

tonysm:
### Added

- Adds a PGVector service to make it easy to play with embeddings

---

close #342

### DIFF
--- a/app/Services/PGVector.php
+++ b/app/Services/PGVector.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace App\Services;
+
+class PGVector extends BaseService
+{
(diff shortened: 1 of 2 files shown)


mattstauffer (approved):
Looks good!

PR #351, opened 2024-11-16
Adds Ollama and OpenWebUI services as tools

tonysm:
### Changelog

- Adds Ollama as a tool
- Adds OpenWebUI as a tool

---

These both tools should work together. But we can use Ollama without the OpenWebUI tool (in case we're using Ollama in an application).

To run the Ollama tool (use defaults):

```bash
php ./takeout enable ollama
```

To run the OpenWebUI (use defaults):

```bash
php ./takeout enable openwebui
```

This should start Ollama at port `11434` and OpenWebUI at port `3000`. 

Open [http://localhost:3000](http://localhost:3000) (it takes a while for OpenWebUI to load), navigate to the top-right corner Admin Panel menu:

<details>
<summary>Screenshot</summary>

image

</details>

Then, to the Settings -> Models section, choose a model (like `llama3`), then hit the download button... this will take a while...

<details>
<summary>Screenshot</summary>

image

</details>

Once it's done downloading, you can hit "New Chat" at the top-left corner, then choose the downloaded model in the model dropdown, then use it

<details>
<summary>Screenshots</summary>

image

</details>

### DIFF
--- a/app/Services/Ollama.php
+++ b/app/Services/Ollama.php
@@ -36,7 +36,7 @@ class Ollama extends BaseService
         parent::prompts();
 
         if (! in_array(strtolower(trim($this->promptResponses['use_gpu'])), ['no', 'n', '0', 'false', ''])) {
-            $this->dockerRunTemplate = '--gpus=all '.$this->dockerRunTemplate;
+            $this->dockerRunTemplate = '--gpus=all ' . $this->dockerRunTemplate;
         }
     }
 }
--- a/app/Services/Ollama.php
+++ b/app/Services/Ollama.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace App\Services;
+
+class Ollama extends BaseService
+{
+    protected static $category = Category::TOOLS;
+
+    protected $organization = 'ollama';
+
+    protected $imageName = 'ollama';
+
+    protected $defaultPort = 11434;
+
+    protected $prompts = [
+        [
+            'shortname' => 'volume',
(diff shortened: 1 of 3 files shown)


mattstauffer (approved):

app/Services/Ollama.php
```diff
@@ -0,0 +1,42 @@
...
+
+    protected $prompts = [
+        [
+            'shortname' => 'volume',
+            'prompt' => 'Volume Name (Ollama will store the models locally)',
+            'default' => 'ollama_data',
+        ],
+        [
+            'shortname' => 'use_gpu',
+            'prompt' => 'Do want to GPUs? Docker on Mac doesn\'t support GPUs. Default is "no", meaning CPU-only. To use GPUs, answer "yes".',
```
mattstauffer, line 23: "Do want to GPUs" maybe means "Do you want to use your GPUs?"

app/Services/OpenWebui.php
```diff
@@ -0,0 +1,54 @@
...
+
+    protected $prompts = [
+        [
+            'shortname' => 'volume',
+            'prompt' => 'Volume Name (to store your app data, like chats, customizations, etc.)',
+            'default' => 'openwebui_data',
+        ],
+        [
+            'shortname' => 'use_gpu',
+            'prompt' => 'Do want to GPUs? Docker on Mac doesn\'t support GPUs. Default is "no", meaning CPU-only. To use GPUs, answer "yes".',
```
mattstauffer, line 29: Same note here re GPUs

PR #382, opened 2025-07-04
Change the Microsoft SQL Server base image

### LINKED ISSUE #380: Retirement: Azure SQL Edge will be retired on September 30th, 2025
I noticed `takeout enable mssql` uses `azure-sql-edge`. This version will retire by the end of September.
Will "SQL Server Express edition" be used in the future by any chance?

https://azure.microsoft.com/en-us/updates?id=azure-sql-edge-retirement

Thanks!

tonysm:
### Changed

- The `azure-sql-edge` is going to be retired later this year. We need to switch to the Express edition of the base image. That one, however, doesn't provide a native ARM64 build, so we need to rely on Rosetta for it. To avoid warnings, we're specifying a platform (can be anyone that works, I'm choosing `linux/amd64`) so Rosetta can make its magic.
- Extracted the Platform detection to an external class and made it _fakeable_ (easier to test), since I needed to check for that outside of the DockerTags class.

---

closes #380


tonysm:
There are some style issues here, but I'll run [Duster](https://github.com/tighten/duster) on the codebase soon.

PR #384, opened 2025-12-10
Update PostgreSql.php for v18

williamjulianvicary:
V18 of Postgres requires a different path for data storage. This PR drops the "data" from the path and moves the volume to the `/var/lib/postgresql` path instead.

See thread for details.
https://github.com/docker-library/postgres/pull/1259

Meanwhile for anyone seeing this, if you want a quick fix to spin up Postgres you can just do this:
`docker run --name Postgres -p 5432:5432 -e POSTGRES_PASSWORD=password -d postgres`

Adjusting the parameters accordingly.


williamjulianvicary:
CI failing seems to be unrelated, presumably due to the underlying docker version on latest being bumped to 18?

